PR #3, opened 2021-03-10
Support for newer Bison versions

sleepyeldrazi:
Updated deprecated code to work with bison 3.0.4 and newer.


curufinwe:
Please also move ArithmeticExpressionParser.in to ..Parser.yy and remove the copy operation from the Makefile. Thanks.

curufinwe (approved):

PR #5, opened 2021-04-09
Singularity definiton files for ARM

sleepyeldrazi:
Created singularity container definition files for ARM, based on the original definition files.


curufinwe:
rasr-arm.def reference tensorflow.sif, it should be tensorflow-arm.sif

curufinwe (approved):

PR #7, opened 2021-07-09
deleted feature vectors copying in setLookAhead()

Tanzp01:
In DelayedRecognizer.cc file, the feature vector copying in the RecognizerDelayHandler::setLookAhead() function took approximately 10% of the time of RecognizerDelayHandler::flush(), which turned out to be unnecessary overhead. 

So this copying part was deleted and featureBuffer_ is now passed directly, where the function signatures were also adapted.

### DIFF
--- a/src/Search/AdvancedTreeSearch/AcousticLookAhead.cc
+++ b/src/Search/AdvancedTreeSearch/AcousticLookAhead.cc
@@ -664,18 +664,18 @@ void AcousticLookAhead::getSuccessorMixtures(const PersistentStateTree&         
     }
 }
 
-void AcousticLookAhead::setLookAhead(std::vector<Mm::FeatureVector> lookahead) {
+void AcousticLookAhead::setLookAhead(std::deque<Core::Ref<const Mm::Feature>> const& lookahead) {
+    
     if (!acousticLookAheadScorer_)
         return;
 
-    if (lookahead.size() > acousticLookaheadDepth_)
-        lookahead.resize(acousticLookaheadDepth_);
+    size_t lookahead_size = std::min(int(lookahead.size()), int(acousticLookaheadDepth_));
 
     acousticLookAhead_.clear();
 
-    for (u32 a = 0; a < lookahead.size(); ++a) {
-        acousticLookAhead_.push_back(std::make_pair(acousticLookAheadScorer_->multiplyAndQuantize(lookahead.at(a))[0],
-                                                    acousticLookAheadScorer_->getScorer(lookahead.at(a))));
+    for (u32 a = 0; a < lookahead_size; ++a) {
+        acousticLookAhead_.push_back(std::make_pair(acousticLookAheadScorer_->multiplyAndQuantize(*(*lookahead.at(a)).mainStream())[0],
+                                                    acousticLookAheadScorer_->getScorer(*(*lookahead.at(a)).mainStream())));
         verify(acousticLookAhead_[a].first.size());
     }
 }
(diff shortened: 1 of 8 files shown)


src/Search/AdvancedTreeSearch/AcousticLookAhead.cc
```diff
@@ -664,18 +664,18 @@ void AcousticLookAhead::getSuccessorMixtures(const PersistentStateTree&
...
-        lookahead.resize(acousticLookaheadDepth_);
+    size_t lookahead_size = std::min(int(lookahead.size()), int(acousticLookaheadDepth_));
 
     acousticLookAhead_.clear();
 
-    for (u32 a = 0; a < lookahead.size(); ++a) {
-        acousticLookAhead_.push_back(std::make_pair(acousticLookAheadScorer_->multiplyAndQuantize(lookahead.at(a))[0],
-                                                    acousticLookAheadScorer_->getScorer(lookahead.at(a))));
+    for (u32 a = 0; a < lookahead_size; ++a) {
+        acousticLookAhead_.push_back(std::make_pair(acousticLookAheadScorer_->multiplyAndQuantize(*(*lookahead.at(a)).mainStream())[0],
```
curufinwe, line 677: instead of `*(x).` use `x->`

src/Search/AdvancedTreeSearch/AdvancedTreeSearch.hh
```diff
@@ -27,6 +27,8 @@
 #include <Speech/ModelCombination.hh>
 #include "DynamicBeamPruningStrategy.hh"
 #include "Trace.hh"
+#include <Mm/Types.hh> // include this so that the Mm::Feature type in setLookAhead() function can be recognized
```
curufinwe, line 30: move this include to the other includes with `<>` and sort it in alphabetically. Also: comment not needed here.

src/Search/Search.hh
```diff
@@ -126,7 +126,7 @@ public:
      * as returned by lookaheadLength().  At the end of a segment, less than requested,
      * or even zero feature vectors may be given.
      */
-    virtual void setLookAhead(const std::vector<Mm::FeatureVector>&) {}
+    virtual void setLookAhead(const std::deque<Core::Ref<const Speech::Feature>>&) {}
```
curufinwe, line 129: Also use `Mm::Feature` here instead of `Speech::Feature`

curufinwe (changes requested):

src/Search/Search.hh
Tanzp01: I tried Mm::Feature at first, but the compiler did not recognize Mm::Feature as Speech::Feature and I got this compiler error: `no known conversion for argument 1 from ‘Speech::RecognizerDelayHandler::FeatureBuffer {aka std::deque<Core::Ref<const Speech::Feature> >}’ to ‘const std::deque<Core::Ref<const Mm::Feature> >&’`.
Static_cast does not work here either.

Tanzp01 (commented):
corrected the bugs and format problem in AcousticLookAhead.cc and AdvancedTreeSearch.hh.

src/Search/Search.hh
curufinwe: OK, but then the lookahead code is never called in the other cases. I.e. this function here is not pure virtual, so if the other classes use a different signature these are new functions and thus never get called. Then change all of them to Speach::Feature.

Tanzp01 (commented):
changed all the corresponding Mm::Feature to Speech::Feature

curufinwe (approved):

SimBe195 (approved):

PR #9, opened 2021-11-08
FFMPEG channel-layout bug fix

sleepyeldrazi:
Fixed FFMPEG bug in order to correctly resample WAV.


curufinwe (changes requested):
Code looks good in general, but please replace tabs with spaces. Thanks!

curufinwe (approved):

PR #10, opened 2021-12-06
Make VanillaGraphLoader work with SavedModel files (.pb)

mattiadg:


curufinwe (changes requested):
Only had a very rough look through the PR. But 2 main points:
1. setSession should not be acessible to everyone, maybe make this available to friend classes only (I want to avoid exposing the tensorflow API outside the Tensorflow module in RASR). This is mainly to protect us from having to change too many places if Google decides to change the API
2. Do not modify the vanilla graph loader, please create a new one

mattiadg:
Do I need to add somehow the new class to the graph-loader options or it is done automatically?

curufinwe:
Yes, you need to add it here: https://github.com/rwth-i6/rasr/blob/master/src/Tensorflow/Module.cc#L33

mattiadg:
Maybe you didn't receive a notification, but I added the factory call

curufinwe:
No longer needed.

PR #14, opened 2022-04-13
Allophone State Fsa bug for arc's weight leaving silence 

Marvin84:
Given a static allophone Finite State Acceptor (FSA), the `ClassicTransducerBuilder`, applies the transition penalties on the arcs. The Fsa at this level makes use of Epsilon arcs in order to distinguish between different word ends, including silence. In the current implementation, any arc coming after Epsilon arc will have speech forward penalty. This means that after trimming the Epsilon arcs, the transition penalty for leaving silence and final speech state is always the respective exit penalties plus speech forward. This should be corrected in the `doExit()` function in order to subtract the speech forward and add the silence forward, when leaving silence.

### DIFF
--- a/src/Am/ClassicTransducerBuilder.cc
+++ b/src/Am/ClassicTransducerBuilder.cc
@@ -784,8 +784,8 @@ typedef std::unordered_map<MinimizedState, Fsa::State*, MinimizedState::Hash,
  *  phone1	   the second state of a phone segment
  *
  * silence.exit = silence.forward - phone?.forward due to epsilon arcs. This happens in the legacy
- * StandardApplicator that is buggy, and has been corrector by implementing
- * ApplicatorWithContext, where the weights are read by refering to the correct source state
+ * LegacyApplicator that is buggy, and has been corrector by implementing
+ * CorrectedApplicator, where the weights are read by refering to the correct source state
  * in case we have an epsilon arc
  **/
 Fsa::ConstAutomatonRef ClassicTransducerBuilder::createEmissionLoopTransducer(
--- a/src/Am/ClassicTransducerBuilder.cc
+++ b/src/Am/ClassicTransducerBuilder.cc
@@ -137,6 +137,12 @@ static const Core::ParameterBool paramFixAllophoneContextAtWordBoundaries(
         "Enabling this option would fix that.",
         false);
 
+static const Core::ParameterBool paramFixTdpLeavingEpsilonArc(
+        "fix-tdp-leaving-epsilon-arc",
+        "Calls a different applicator that keeps track of the emission label at previous step "
+        "This solved the inconsistent transition model application due to epsilon arcs ",
+        false);
+
 ClassicTransducerBuilder::ClassicTransducerBuilder(Core::Ref<
                                                    const ClassicAcousticModel>
                                                            model)
@@ -153,6 +159,7 @@ ClassicTransducerBuilder::ClassicTransducerBuilder(Core::Ref<
     inputType_                              = inputAcceptsEmissionLabels;
     filterOutInvalidAllophones_             = paramFilterOutInvalidAllophones(model->getConfiguration());
     fixAllophoneContextAtWordBoundaries_    = paramFixAllophoneContextAtWordBoundaries(model_->getConfiguration());
+    fixTdpLeavingEpsilonArc_                = paramFixTdpLeavingEpsilonArc(model_->getConfiguration());
     statistics_                             = new Statistics;
 }
 
@@ -644,7 +651,13 @@ Fsa::ConstAutomatonRef ClassicTransducerBuilder::applyTransitionModel(
     else if (ff->inputAlphabet() == emissions_)
         silenceLabel = model_->stateTyingRef_->classify(silenceState);
 
+    if (fixTdpLeavingEpsilonArc_){
+        return model_->transitionModel_->applyWithContext(ff, silenceLabel, (nDisambiguators_ == 0));
+    }
+
     return model_->transitionModel_->apply(ff, silenceLabel, (nDisambiguators_ == 0));
+
+
 }
 
 void ClassicTransducerBuilder::applyStateTying() {
--- a/src/Am/ClassicTransducerBuilder.cc
+++ b/src/Am/ClassicTransducerBuilder.cc
@@ -783,8 +783,8 @@ typedef std::unordered_map<MinimizedState, Fsa::State*, MinimizedState::Hash,
  *  phone0	   the first state of a phone segment
  *  phone1	   the second state of a phone segment
  *
- * silence.exit = silence.forward - phone?.forward due to epsilon arcs. This happens in the s
- * StandardApplicator that is a legacy buggy applicator, and has been corrector by implementing
+ * silence.exit = silence.forward - phone?.forward due to epsilon arcs. This happens in the legacy
+ * StandardApplicator that is buggy, and has been corrector by implementing
  * ApplicatorWithContext, where the weights are read by refering to the correct source state
  * in case we have an epsilon arc
  **/
--- a/src/Am/ClassicTransducerBuilder.cc
+++ b/src/Am/ClassicTransducerBuilder.cc
@@ -783,7 +783,7 @@ typedef std::unordered_map<MinimizedState, Fsa::State*, MinimizedState::Hash,
  *  phone0	   the first state of a phone segment
  *  phone1	   the second state of a phone segment
  *
- * silence.exit = silence.forward - phone?.forward   but why?
+ * silence.exit = silence.forward - phone?.forward due to epsilon arcs. This is corrected in doExit() funtion.
  **/
 Fsa::ConstAutomatonRef ClassicTransducerBuilder::createEmissionLoopTransducer(
         bool transitionModel) {
--- a/src/Am/ClassicTransducerBuilder.cc
+++ b/src/Am/ClassicTransducerBuilder.cc
@@ -783,7 +783,10 @@ typedef std::unordered_map<MinimizedState, Fsa::State*, MinimizedState::Hash,
  *  phone0	   the first state of a phone segment
  *  phone1	   the second state of a phone segment
  *
- * silence.exit = silence.forward - phone?.forward due to epsilon arcs. This is corrected in doExit() funtion.
+ * silence.exit = silence.forward - phone?.forward due to epsilon arcs. This happens in the s
+ * StandardApplicator that is a legacy buggy applicator, and has been corrector by implementing
+ * ApplicatorWithContext, where the weights are read by refering to the correct source state
+ * in case we have an epsilon arc
  **/
 Fsa::ConstAutomatonRef ClassicTransducerBuilder::createEmissionLoopTransducer(
         bool transitionModel) {
--- a/src/Am/ClassicTransducerBuilder.cc
+++ b/src/Am/ClassicTransducerBuilder.cc
@@ -137,12 +137,6 @@ static const Core::ParameterBool paramFixAllophoneContextAtWordBoundaries(
         "Enabling this option would fix that.",
         false);
 
-static const Core::ParameterBool paramFixTdpLeavingEpsilonArc(
-        "fix-tdp-leaving-epsilon-arc",
-        "Calls a different applicator that keeps track of the emission label at previous step "
-        "This solved the inconsistent transition model application due to epsilon arcs ",
-        false);
-
 ClassicTransducerBuilder::ClassicTransducerBuilder(Core::Ref<
                                                    const ClassicAcousticModel>
                                                            model)
@@ -159,7 +153,6 @@ ClassicTransducerBuilder::ClassicTransducerBuilder(Core::Ref<
     inputType_                              = inputAcceptsEmissionLabels;
     filterOutInvalidAllophones_             = paramFilterOutInvalidAllophones(model->getConfiguration());
     fixAllophoneContextAtWordBoundaries_    = paramFixAllophoneContextAtWordBoundaries(model_->getConfiguration());
-    fixTdpLeavingEpsilonArc_                = paramFixTdpLeavingEpsilonArc(model_->getConfiguration());
     statistics_                             = new Statistics;
 }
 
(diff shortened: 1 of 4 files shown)


curufinwe:
We don't use this much anymore, but does this code also work with skip connections? What if we have state-repetition set to 2 (because you work with phon0 here, but there is also phon1)?

Marvin84:
>

Marvin84:
> What if we have state-repetition set to 2 (because you work with phon0 here, but there is also phon1)?
As far as I see here, this is a transition-specific state type, and is not about the distinction between different state labels but just speech and silence. There is actually no use of `phone1` in any part of the code. However, for a cleaner solution, I will just use the `current.weights` which should have the correct state type.

Marvin84:
> but does this code also work with skip connections?

Good point, this should be also corrected for the skip, where we have a similar problem. I will do the changes.

albertz:
Should we introduce an option to enable the new fixed behavior? I assume people who upgrade RASR would not expect that the behavior suddenly is different from before.

In RETURNN, we introduced the "behavior version" (https://github.com/rwth-i6/returnn/issues/508) such that the default settings can be reasonable when people use a recent behavior version while not breaking any older setup which uses an older behavior version. I think RASR doesn't have such a concept yet. Not sure if we want to introduce it (but this should maybe be a separate discussion).

In any case, I think the behavior should not just suddenly change, even if the old behavior was somehow wrong. For other things we also have always introduced options (@Marvin84 remember the other options regarding FSA stuff).

Marvin84:
> the behavior should not just suddenly change

Here, in my opinion, we are not dealing with a behavior change due to the introduction of a new feature or functionality, but with a pure algorithmic bug. Surely, everybody should be notified about this, but continuing to use the 'old behavior' means applying a transition model that is incorrect.

albertz:
> ... but continuing to use the 'old behavior' means applying a transition model that is incorrect.

Yes. But we also have this case for some of the other options I introduced before.

"incorrect" is maybe also a strong word. Maybe "unexpected" is better. Because it "worked" already before (since many years).

People who would update RASR now suddenly might even get worse results when they don't check and maybe adapt the config?

Marvin84:
> "incorrect" is maybe also a strong word. Maybe "unexpected" is better.

leaving silence state with probability of speech forward is not unexpected is just incorrect

Marvin84:
> People who would update RASR now suddenly might even get worse results when they don't check and maybe adapt the config?

Yes, this should be tested. But this is independent of fixing the bug.

albertz:
> > People who would update RASR now suddenly might even get worse results when they don't check and maybe adapt the config?
> 
> Yes, this should be tested. But this is independent of fixing the bug.

Well, I personally would not introduce some new changed behavior (even when the old behavior was "wrong") without having an explicit option to enable the new behavior.

At least in RETURNN, we are very strict about never breaking old setups or changing the behavior of old setups. I think this is a good policy in general, also for RASR. But maybe @curufinwe, @michelwi, or also @ZhouW321 have different opinions on this? @JackTemaki maybe?

src/Am/TransitionModel.cc
```diff
@@ -532,11 +532,25 @@ class TransitionModel::Applicator {
     void doExit(const StackItem& current, const Fsa::Arc* ra) {
         require(alphabet_->isDisambiguator(ra->input()));
         verify(!applyExitTransitionToFinalStates_);
+        Fsa::Weight correctedWeight;
+        if (current.emission == silenceLabel_){
+            //we need to correct for the algorithmic bug introduced by epsilon arcs in case of exit,
```
albertz, line 537: The comment-style is inconsistent to the remaining comment-style in this file. I think there should be a space after `//`.

src/Am/TransitionModel.cc
```diff
@@ -532,11 +532,25 @@ class TransitionModel::Applicator {
...
+            //where independently from the source state identity (speech/silence)
+            //the speech forward penalty is applied. Please note that leaving silence is conceptually
+            //exiting and forwarding from silence, we can never only forward from silence.
+            StateType speechType = TransitionModel::phone0;
+            StateType silType    = TransitionModel::silence;
+            Fsa::Weight speechFwdToSubtract = t_->semiring()->invert(Fsa::Weight((*transitionModel_[speechType])[StateTransitionModel::forward]));
+            Fsa::Weight correctedTdp = t_->semiring()->extend(ra->weight(),  speechFwdToSubtract);
+            correctedWeight = t_->semiring()->extend(correctedTdp, Fsa::Weight((*transitionModel_[silType])[StateTransitionModel::forward]));
```
albertz, line 545: I wonder if this is the best place to have this. (It's long ago that I worked with this code so I don't really remember anymore.) For example, can't we directly do this when we introduce the eps arcs? It's sounds a bit like we need this here as a workaround for another bug.

Marvin84: Yes, this solution is kinf od hacky, because it corrects for this specific algorithmic solution and this specific fsa.
The clean solution would be to extend the `State` struct to have also the information about the emission label of the previous arc. But this would be even a bigger change and would actually need probably a separate module.

Marvin84: > because you work with phon0 here

I just saw that the `statetype(Fsa::LabelId emission)` function here in this class also distinguishes only between silence and `phone0`.

Marvin84:
Hi everybody, 
Following @albertz 's suggestion I am proposing a less hacky version, where a proper fsa implementation with context is provided. Furthermore, I introduced a flag that will decide between the default and the new behavior.
Looking forwrad to your suggestions.

JackTemaki:
In my opinion, as we were previously not using this particular version of RASR in experiments, I think it is fine to directly make the new behavior the default.

albertz:
> In my opinion, as we were previously not using this particular version of RASR in experiments, I think it is fine to directly make the new behavior the default.

I don't understand this argument. We would expect that all versions of RASR behave the same, or not? Otherwise it complicates everything to switch over to another RASR versions.

JackTemaki:
> > In my opinion, as we were previously not using this particular version of RASR in experiments, I think it is fine to directly make the new behavior the default.
> 
> I don't understand this argument. We would expect that all versions of RASR behave the same, or not? Otherwise it complicates everything to switch over to another RASR versions.

I do not consider them behaving the same. It only complicates anything it if this affects not only full-sum training. Because in that case we anyway have multiple different versions of RASR. Also changes to RASR are very rare (the last algorithmic change was more than one year ago), So we could even just start (or rather continue) giving version numbers for RASR itself.

Marvin84:
I already changed the first implementation proposed in the original pull request with the option of having a flag that selects new or old behavior. I would appreciate if you could give feedback on the specific implementation. @curufinwe @JackTemaki

src/Am/ClassicTransducerBuilder.cc
```diff
@@ -137,6 +137,12 @@ static const Core::ParameterBool paramFixAllophoneContextAtWordBoundaries(
         "Enabling this option would fix that.",
         false);
 
+static const Core::ParameterBool paramFixTdpLeavingEpsilonArc(
+        "fix-tdp-leaving-epsilon-arc",
+        "Calls a different applicator that keeps track of the emission label at previous step "
+        "This solved the inconsistent transition model application due to epsilon arcs ",
+        false);
```
JackTemaki, line 144: What I meant in my comments is changing this to true.

albertz: I very much disagree here. I think there should never be behavior changes between RASR versions. Users should be able to switch or update RASR versions without being afraid that this would break their setup.

I have no idea how you get to this idea that it should be the other way around.

Marvin84: I think I already mentioned that the old behavior is incorrect on the theoretical level. We already tested the new code and in practice we get the same WER. I am fine with having this parameter false, but it should be clear that the default behavior does not correspond to a proper transition model when leaving silence states.

JackTemaki: > I have no idea how you get to this idea that it should be the other way around

Again, I am stating in my opinion we should (in this specific case) change this flag to default `True` as:
 - It does not affect our regular hybrid training (if it does in a relevant manner, then surely it should be False)
 - It is (as Tina said) definitely a wrong implementation, not only a different, so I would consider this a bug-fix and not a behavioural change.
 - I do not know of anyone using the Github master branch of RASR in experiments, and we are currently in the process of making the master usable for i6, so I would want the correct behavior as default when I switch.
 - We have multiple incompatible RASR versions, and the master branch here will be the one where we try to unify at least some parts of them.

albertz: > It does not affect our regular hybrid training

I thought it does affect realignments, full-sum training, and maybe also search?

> It is (as Tina said) definitely a wrong implementation, not only a different, so I would consider this a bug-fix and not a behavioural change.

A change in behavior is a change in behavior.

> I do not know of anyone using the Github master branch of RASR in experiments, and we are currently in the process of making the master usable for i6, so I would want the correct behavior as default when I switch.

I don't understand. I would want that when I switch RASR versions that my setups behave as before, even if it was incorrect before.

> We have multiple incompatible RASR versions, and the master branch here will be the one where we try to unify at least some parts of them.

Yes, and this makes it even more important that people can just switch to the master branch and don't have to expect that their setups will suddenly work differently.

But anyway, I think this is a decision for @curufinwe.

In RETURNN, we definitely also handle it like that, that even if there are bugs, we introduce a new behavior version, such that people can always safely stay on master branch.

src/Am/TransitionModel.cc
```diff
@@ -641,6 +641,431 @@ Fsa::ConstAutomatonRef TransitionModel::apply(Fsa::ConstAutomatonRef in,
     return ap.apply(in);
 }
 
+
+
+class TransitionModel::ApplicatorWithContext {
```
curufinwe, line 646: There is a lot of duplication of code in this class. Please refactor this into a common base class and have Applicator and ApplicatorWithContext inherit from it.

src/Am/TransitionModel.hh
```diff
@@ -121,6 +122,11 @@ public:
             Fsa::LabelId           silenceLabel,
             bool                   applyExitTransitionToFinalStates) const;
 
+    Fsa::ConstAutomatonRef applyWithContext(
```
curufinwe, line 125: Add documentation describing how this function is different from apply.

curufinwe (changes requested):

src/Am/ClassicTransducerBuilder.cc
JackTemaki: I know that the default batch-norm settings are far from "correct" in the sense of what you would expect, and this was fixed with a new "behavior" version, but I don't know of anything else.

Marvin84: I took a quick look at the link above. In my opinion, these are all real behavior changes. Also the case of batch normalization it seems to me that this is more of a different implementation choice, involving hyperparameters and aspects regarding complexity optimization. How is this comparable with our example here where we have a conceptually wrong modeling approach, i.e. we have an automaton that has the probability of leaving silence equal to the probability of leaving speech plus exiting silence.

albertz: It's exactly the same. In all cases, the behavior was wrong before and is now correct. E.g. the batch norm settings before were really totally wrong, conceptually wrong, did not make any sense. The other examples are all similar.

You could also call this option you introduce here just a hyper param.

Marvin84: I see it differently. But I think there is no point in continuing this discussion. The real issue we need to discuss is whether in rasr we can consider similar behavior-monitoring as in returnn. And this should solve the issue.

src/Am/TransitionModel.cc
curufinwe: Yes

Marvin84:
With the help of @curufinwe the architecture refactoring following the implemented logic is carried out.
@SimBe195 @JackTemaki @curufinwe the current code has been already tested. Please add your review.

src/Am/TransitionModel.cc
```diff
@@ -629,16 +597,227 @@ Fsa::ConstAutomatonRef TransitionModel::Applicator::apply(Fsa::ConstAutomatonRef
...
+    Core::Choice::Value appChoice = Am::TransitionModel::paramApplicatorType(config);
+    if (appChoice == Core::Choice::IllegalValue)
+        criticalError("Unknwon Transition Applicator type.");
+
+    switch (appChoice) {
+        case LegacyBuggyApplicator:
+            ap.reset(new StandardApplicator(*this));
+            break;
+        case CorrectedApplicator:
+            ap.reset(new ApplicatorWithContext(*this));
```
SimBe195, line 810: I don't really like that the `ApplicatorType`s refer to classes which are named differently and have names which themselves sound like subclasses of `AbstractApplicator`. I would perhaps somewhat unify the naming scheme like this:

- `StandardApplicator` -> `LegacyApplicator`
- `LegacyBuggyApplicator` -> `Legacy` or `LegacyType`
- `"legacy-buggy"` -> `"legacy"`
- `ApplicatorWithContext` -> `CorrectedApplicator`
- `CorrectedApplicator` -> `Corrected` or `CorrectedType`

That way it's clearer which ones belong together.

SimBe195 (changes requested):

curufinwe (changes requested):

src/Am/TransitionModel.hh
```diff
@@ -77,20 +77,23 @@ public:
         phone1,
         nStateTypes
     };
-    class Applicator;
+    enum TyingType {
```
curufinwe, line 80: These are not scoped enums. Scoped enums are defined by adding either the `class` or `struct` keyword after the `enum` keyword. This is why you could not name the enum values LegacyApplicator and CorrectedApplicator as in an unscoped enum these names are added to the surrounding namespace which already contains these names for the classes.

Marvin84: Could we please open another pull request for this? This does not have any effect on the config side and can be fixed later.

Marvin84:
@curufinwe and @SimBe195 I reversed for now to the classic enum. I will then open another pull request for this to have a consistent solution for all enums under `TransitionModel`. Could you please merge now?

SimBe195 (approved):

curufinwe (approved):

PR #15, opened 2022-05-25
generate Modules.hh before starting parallel build

JackTemaki:
Otherwise compilation will crash with `Modules.hh` not found, as it does not exist yet when building parallel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,7 @@ world:
 
 all:	build test
 
-build:	config
-	$(MAKE) -C src Modules.hh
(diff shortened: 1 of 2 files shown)


michelwi:
> > you should also move the lines that indicate how to build target Modules.hh i.e. lines 127-134
> 
> Move them where?

sorry, I meant from `src/Marefile` to `Makefile` just like how you moved the make target `Modules.hh`

michelwi (approved):

SimBe195 (approved):

PR #16, opened 2022-05-27
simple history LM

ZhouW321:
baseLM for next-step simple TF NLM
- A simple language model for token history extension and hashing
- also useful for no-LM recognition but still with recombination capability  or  just separate recombineLM usage

### DIFF
--- a/src/Lm/SimpleHistoryLm.hh
+++ b/src/Lm/SimpleHistoryLm.hh
@@ -7,7 +7,7 @@
 namespace Lm {
 
   /**
-   * A simple language model for token history extension and hashing (so far full history)
+   * A simple language model for token history extension and hashing (recombination)
    * also useful for no-LM recognition but still with recombination capability
    */
 
@@ -16,6 +16,7 @@ namespace Lm {
     TokenIdSequence tokIdSeq;
     mutable u32 refCount;
 
+    SimpleHistory(): refCount(0) {}
     SimpleHistory(Bliss::Token::Id tid): tokIdSeq(1, tid), refCount(0) {}
     SimpleHistory(const TokenIdSequence& r, Bliss::Token::Id tid): tokIdSeq(r), refCount(0) { tokIdSeq.push_back(tid); }
   };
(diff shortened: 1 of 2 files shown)


src/Lm/SimpleHistoryLm.hh
```diff
@@ -0,0 +1,100 @@
...
+
+      History extendedHistory(const History& h, Token w) const
+      {
+        const SimpleHistory* sh = static_cast<const SimpleHistory*>(h.handle());
+        SimpleHistory* nsh = new SimpleHistory(sh->tokIdSeq, w->id());
+        return history(nsh);
+      }
+
+      // TODO reduced history for limited context
+      History reducedHistory(const History& h, u32 limit) const { return h; }
```
JackTemaki, line 83: Regarding the TODO: Why would you need to implement this, if the score is anyway always zero? Is this something the implementation needs to provide for some future usage?

src/Lm/SimpleHistoryLm.hh
```diff
@@ -0,0 +1,100 @@
...
+      {
+        const SimpleHistory* sh = static_cast<const SimpleHistory*>(handle);
+        return token_id_sequence_hash(sh->tokIdSeq);
+      }
+
+      bool isEquivalent(HistoryHandle lhd, HistoryHandle rhd) const
+      { // lhd != rhd when reaching here
+        const SimpleHistory* lsh = static_cast<const SimpleHistory*>(lhd);
+        const SimpleHistory* rsh = static_cast<const SimpleHistory*>(rhd);
+        //return token_id_sequence_hash(lsh->tokIdSeq) == token_id_sequence_hash(rsh->tokIdSeq);
```
JackTemaki, line 54: Can this be removed, or is this something that could in theory work later on?

src/Lm/SimpleHistoryLm.hh
ZhouW321: might be needed for recombination based on limited history

src/Lm/SimpleHistoryLm.hh
ZhouW321: well, both should work, I think the commented one is safer, but the one below is faster.

PR #17, opened 2022-05-30
Add ./configure script

JackTemaki:
I added a script which we can use to change the most used compiled options in RASR, which also has an `--i6` flag to directly configure RASR for a default i6 build.

This is just a draft, but I want your opinions on this.

### DIFF
--- a/configure
+++ b/configure
@@ -45,10 +45,10 @@ set_python () {
     if [ $MINOR -ge 8 ]; then
         # python version beginning from 3.8 do not include -lpython3.X when calling
         # python3-config without --embed, so add it if such minor version is found
-        grep "python3-config --libs --embed" config/os-${OS}.make > /dev/null
+        grep "python3-config --ldflags --embed" config/os-${OS}.make > /dev/null
         if [ $? -ne 0 ]; then
             echo "Python minor version is > 8, add --embed"
-            sed -i "s/\\(python3-config --libs\\)/\1 --embed/g" config/os-${OS}.make
+            sed -i "s/\\(python3-config --ldflags\\)/\1 --embed/g" config/os-${OS}.make
         else
             echo "Python3: --embed already set"
         fi
--- a/configure
+++ b/configure
@@ -138,7 +138,7 @@ Module options:
 
 Dependency options:
   --set-openfst=OPENFST_DIR             set the path to an OpenFST directory
-  --set-python=PYTHON_DIR           set the folder of a python installation
+  --set-python=PYTHON_DIR               set the folder of a python installation
                                         which contains /bin, /lib etc..
   --set-tensorflow=TF_COMPILE_BASE      set the Tensorflow root directory path
 EOF
@@ -165,8 +165,8 @@ for i in "$@"; do
             shift
             ;;
         --set-python=*)
-            PYTHON_BIN_PATH="${i#*=}"
-            set_python $PYTHON_BIN_PATH
+            PYTHON_DIR="${i#*=}"
+            set_python $PYTHON_DIR
             shift
             ;;
         --set-march=*)
--- a/configure
+++ b/configure
@@ -33,7 +33,8 @@ set_openfst () {
 }
 
 set_python () {
-    VERSION_STRING=$($1/python3 --version | grep -o "[0-9.]*")
+    PYTHON_BIN_PATH=$1
+    VERSION_STRING=$(${PYTHON_BIN_PATH}/python3 --version | grep -o "[0-9.]*")
     if [ $? -ne 0 ];then exit 1;fi
     VERSION_ARRAY=( ${VERSION_STRING//./ } )
     MAJOR=${VERSION_ARRAY[0]}
@@ -47,16 +48,13 @@ set_python () {
         # python3-config without --embed, so add it if such minor version is found
         grep "python3-config --ldflags --embed" config/os-${OS}.make > /dev/null
         if [ $? -ne 0 ]; then
-            echo "Python minor version is > 8, add --embed"
+            echo "Python minor version is > 8, adding --embed"
             sed -i "s/\\(python3-config --ldflags\\)/\1 --embed/g" config/os-${OS}.make
-        else
-            echo "Python3: --embed already set"
         fi
     fi
-    PYTHON_BIN_PATH=$1
     # make sure there is a trailing slash
     [[ "${PYTHON_BIN_PATH}" != */ ]] && PYTHON_BIN_PATH="${PYTHON_BIN_PATH}/"
-    set_variable "PYTHON_BIN_PATH" $PYTHON_BIN_PATH
+    set_variable "PYTHON_BIN_PATH" "$PYTHON_BIN_PATH"
 }
 
 set_tensorflow () {
--- a/configure
+++ b/configure
@@ -60,7 +60,7 @@ set_python () {
 }
 
 set_march () {
-    sed -i "s/-march=native/-march=$1/g" config/proc-x86_64.make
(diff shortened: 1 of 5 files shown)


configure
```diff
@@ -0,0 +1,169 @@
...
+    MAJOR=${VERSION_ARRAY[0]}
+    MINOR=${VERSION_ARRAY[1]}
+    if [ $MAJOR -ne 3 ]; then
+        echo "Error: Expected Python version 3, found $MAJOR"
+        exit 1
+    fi
+    if [ $MINOR -ge 8 ]; then
+        # python version beginning from 3.8 do not include -lpython3.X when calling
+        # python3-config without --embed, so add it if such minor version is found
+        grep "python3-config --libs --embed" config/os-${OS}.make > /dev/null
```
michelwi, line 48: Are you sure this will always appear in this order?

configure
```diff
@@ -0,0 +1,169 @@
...
+        if [ $? -ne 0 ]; then
+            echo "Python minor version is > 8, add --embed"
```
michelwi, line 50:
```suggestion
            echo "Python minor version is > 8, adding --embed"
```
just to make sure it is interpreted as an information that it is being added and not as suggestion for the user to do so.

configure
```diff
@@ -0,0 +1,169 @@
...
+            sed -i "s/\\(python3-config --libs\\)/\1 --embed/g" config/os-${OS}.make
+        else
+            echo "Python3: --embed already set"
```
michelwi, line 53: I don't think we need this message

configure
```diff
@@ -0,0 +1,169 @@
...
+        fi
+    fi
+    PYTHON_BIN_PATH=$1
+    # make sure there is a trailing slash
+    [[ "${PYTHON_BIN_PATH}" != */ ]] && PYTHON_BIN_PATH="${PYTHON_BIN_PATH}/"
+    set_variable "PYTHON_BIN_PATH" $PYTHON_BIN_PATH
```
michelwi, line 59:
```suggestion
    set_variable "PYTHON_BIN_PATH" "$PYTHON_BIN_PATH"
```
in case of spaces in the `PYTHON_BIN_PATH`

configure
```diff
@@ -0,0 +1,169 @@
...
+        # python3-config without --embed, so add it if such minor version is found
+        grep "python3-config --libs --embed" config/os-${OS}.make > /dev/null
+        if [ $? -ne 0 ]; then
+            echo "Python minor version is > 8, add --embed"
+            sed -i "s/\\(python3-config --libs\\)/\1 --embed/g" config/os-${OS}.make
+        else
+            echo "Python3: --embed already set"
+        fi
+    fi
+    PYTHON_BIN_PATH=$1
```
michelwi, line 56: this could be moved to the beginning of the function and then replace all other `$1` by `$PYTHON_BIN_PATH`

configure
```diff
@@ -0,0 +1,169 @@
...
+            show_help
+            ;;
+        --i6)
+            configure_i6
+            shift
+            ;;
+        --set-tensorflow=*)
+            TF_DIR="${i#*=}"
+            set_variable "TF_COMPILE_BASE" $TF_DIR
+			shift
```
michelwi, line 135:
```suggestion
            shift
```

configure
```diff
@@ -0,0 +1,169 @@
...
+            ;;
+        --set-march=*)
+            MARCH="${i#*=}"
+            set_march $MARCH
+            shift
+            ;;
+        --enable-module=*)
+            MODULE="${i#*=}"
+            enable_module $MODULE
+			shift
```
michelwi, line 155:
```suggestion
            shift
```

configure
```diff
@@ -0,0 +1,169 @@
...
+            ;;
+		--default)
+			DEFAULT=YES
+            echo "default"
+			shift
+			;;
```
michelwi, line 161: what does this do?

configure
```diff
@@ -0,0 +1,169 @@
...
+    fi
+    PYTHON_BIN_PATH=$1
+    # make sure there is a trailing slash
+    [[ "${PYTHON_BIN_PATH}" != */ ]] && PYTHON_BIN_PATH="${PYTHON_BIN_PATH}/"
+    set_variable "PYTHON_BIN_PATH" $PYTHON_BIN_PATH
+}
+
+set_march () {
+    sed -i "s/-march=native/-march=$1/g" config/proc-x86_64.make
+}
```
michelwi, line 64: if somebody wanted to use the i6 settings by default but change the architecture to XX, then `--i6` is executed first and sets this line to `-march=barcelona` and the subsequent call for `set_march XX` will fail silently because 
```
sed -i "s/-march=native/-march=$1/g" config/proc-x86_64.make
```
will not replace any lines.

michelwi (commented):
this PR also contains https://github.com/rwth-i6/rasr/pull/15

configure
JackTemaki: Okay here is some formatting bug... weird...

configure
JackTemaki: unless someone manually edited the config files, yes... but if you manually change stuff you do not need the `./configure` anyway.

configure
```diff
@@ -0,0 +1,169 @@
...
+
+
+
+enable_module () {
```
michelwi, line 68: maybe we also would like to have a `disable_module` function?

JackTemaki:
@michelwi any idea what to do about the dependency issue? Should I just remove the explicit code here and we keep this open?

michelwi:
> @michelwi any idea what to do about the dependency issue? Should I just remove the explicit code here and we keep this open?

I think the dependency issue is a whole separate problem that _someone_ should fix, which is bigger than just `LM_TFRNN -> TENSORFLOW` (which is rather obvious by the name). I would remove specific code for now and create an issue to fix the general dependency check.

configure
```diff
@@ -0,0 +1,194 @@
...
+  --set-march                           set the target architecture to compile for
+                                        (see: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html)
+
+Module options:
+  --enable-module=<MOD_NAME>            enables module "MODULE_<MOD_NAME>"
+  --disable-module=<MOD_NAME>           disables module "MODULE_<MOD_NAME>"
+
+Dependency options:
+  --set-openfst=OPENFST_DIR             set the path to an OpenFST directory
+  --set-python=PYTHON_DIR           set the folder of a python installation
```
michelwi, line 141:
```suggestion
  --set-python=PYTHON_DIR               set the folder of a python installation
```
consistent spacing

configure
```diff
@@ -0,0 +1,194 @@
...
+            shift
+            ;;
+        --set-openfst=*)
+            OPENFST_DIR="${i#*=}"
+            set_openfst $OPENFST_DIR
+            shift
+            ;;
+        --set-python=*)
+            PYTHON_BIN_PATH="${i#*=}"
+            set_python $PYTHON_BIN_PATH
```
michelwi, line 169:
```suggestion
            PYTHON_DIR="${i#*=}"
            set_python $PYTHON_DIR
```
just consistent variable naming

JackTemaki:
> > @michelwi any idea what to do about the dependency issue? Should I just remove the explicit code here and we keep this open?
> 
> I think the dependency issue is a whole separate problem that _someone_ should fix, which is bigger than just `LM_TFRNN -> TENSORFLOW` (which is rather obvious by the name). I would remove specific code for now and create an issue to fix the general dependency check.

Done

SimBe195 (approved):

michelwi (approved):

PR #18, opened 2022-05-31
Add Cache Manager

JackTemaki:
This adds again the cache manager, so that we can run the github version at i6.


JackTemaki:
@michelwi @SimBe195 

This PR would also be ready now!

michelwi:
> @michelwi @SimBe195
> 
> This PR would also be ready now!

I have no objections, but too low confidence to press accept. If you anyway only copied from phabricator, then "should be fine".

JackTemaki:
> > @michelwi @SimBe195
> > This PR would also be ready now!
> 
> I have no objections, but too low confidence to press accept. If you anyway only copied from phabricator, then "should be fine".

Yes, this is 100% code from Phabricator (except obviously for the additional line in `configure`)

SimBe195 (approved):

PR #24, opened 2022-06-10
fullsum training automaton: RNA/CTC/HMM topology

ZhouW321:
sprint automaton based  from-scratch full-sum training using BW Loss

### DIFF
--- a/src/Nn/AllophoneStateFsaExporter.cc
+++ b/src/Nn/AllophoneStateFsaExporter.cc
@@ -25,18 +25,6 @@
 
 namespace {
 
-struct Edge {
-    Edge()
-            : from(0u), to(0u), emission_idx(0u), weight(0.0f) {}
-    Edge(Fsa::StateId from, Fsa::StateId to, Am::AcousticModel::EmissionIndex emission_idx, float cost)
-            : from(from), to(to), emission_idx(emission_idx), weight(cost) {}
-
-    Fsa::StateId                     from;
-    Fsa::StateId                     to;
-    Am::AcousticModel::EmissionIndex emission_idx;
-    float                            weight;
-};
-
 static void apply_state_map(std::vector<Fsa::StateId> const&             state_map,
                             std::vector<Fsa::StateId>&                   states,
                             std::vector<std::pair<Fsa::StateId, float>>& final_states,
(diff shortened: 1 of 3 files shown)


src/Nn/AllophoneStateFsaExporter.cc
```diff
@@ -190,4 +325,60 @@ AllophoneStateFsaExporter::ExportedAutomaton AllophoneStateFsaExporter::exportFs
     return result;
 }
 
+// Note: pronunciation variants ignored here, i.e. no normalization
```
Marvin84, line 328: Do I get it right that here you overwrite the arc weights with only transition scores?
If yes, you are ignoring both lemma level and lexicon level pronunciation normalizations. This makes the implementation not generic enough.
And the question is: why do you even need to have the HMM topology modified here? We already have an FSA builder module `ClassicTransducerBuilder` that prepares the FSA for full-sum with our standard HMM topology that supports both mono-state and 3-partite. Shouldn't you extend that?

src/Nn/AllophoneStateFsaExporter.cc
```diff
@@ -135,11 +123,115 @@ static void make_single_final_state(std::vector<Fsa::StateId>& states, std::vect
...
+        {
+            Fsa::StateId target = arc->target_;
+            if ( target == sId )
+                continue;
+            u32 depth = getStateDepth(target, stateDepth, fsa) + 1;
+            if ( depth < stateDepth[sId] )
+                stateDepth[sId] = depth;
+        }
+        if ( state->isFinal() )
+            stateDepth[sId] = 0;
```
SimBe195, line 142: Why do we need to iterate over the arcs if the state is final? Shouldn't the for loop be only in the "else" case of `if (state->isFinal() )`

src/Nn/AllophoneStateFsaExporter.cc
```diff
@@ -190,4 +325,60 @@ AllophoneStateFsaExporter::ExportedAutomaton AllophoneStateFsaExporter::exportFs
...
+void AllophoneStateFsaExporter::modifyMinDuration(std::vector<Edge>& edges, std::vector<Fsa::StateId>& states, Core::Ref<const Fsa::StaticAutomaton> automaton, f64 time) const
+{   // expanded sequence should not exceed number of frames
+    u32 expand = minOccur_;
+    if ( time >= 0 ) 
+    {   // if negative time: unknown length, assume always expandable
+        u32 maxLength = time / frameShift_ / reduceFrameFactor_;
+        std::vector<u32> stateDepth(automaton->size(), Core::Type<u32>::max);
+        u32 seqLength = getStateDepth(0, stateDepth, automaton); // shortest sequence length
+        while ( seqLength * expand > maxLength && expand > 1 )
+            --expand; // reduce expansion
```
SimBe195, line 358: I would say it's unexpected behavior that the minimum duration may be reduced automatically. In my opinion the user should receive a warning when that happens.

SimBe195 (changes requested):

ZhouW321:
no need anymore

PR #25, opened 2022-06-10
flexible alignment: auto-subsample, peaky, single-state, as-is, etc.

ZhouW321:
also improve code logic of AligningFeatureExtractor

### DIFF
--- a/src/Nn/BufferedAlignedFeatureProcessor.cc
+++ b/src/Nn/BufferedAlignedFeatureProcessor.cc
@@ -30,15 +30,19 @@ using namespace Nn;
 
 template<typename T>
 const Core::ParameterFloat BufferedAlignedFeatureProcessor<T>::paramSilenceWeight(
-        "silence-weight", "weight for silence state", 1.0);
+    "silence-weight", "weight for silence state", 1.0);
 
 template<typename T>
 const Core::ParameterString BufferedAlignedFeatureProcessor<T>::paramClassWeightsFile(
-        "class-weights-file", "file with class-weights-vector");
+    "class-weights-file", "file with class-weights-vector");
 
 template<typename T>
 const Core::ParameterBool BufferedAlignedFeatureProcessor<T>::paramWeightedAlignment(
-        "weighted-alignment", "use weights from alignment", false);
+    "weighted-alignment", "use weights from alignment", false);
+
+template<typename T>
+const Core::ParameterInt BufferedAlignedFeatureProcessor<T>::paramReduceAlignmentFactor(
+    "reduce-alignment-factor", "downsample alignment (only for peaky alignment)", 1);
 
 template<typename T>
 BufferedAlignedFeatureProcessor<T>::BufferedAlignedFeatureProcessor(const Core::Configuration& config, bool loadFromFile)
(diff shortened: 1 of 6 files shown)


SimBe195:
I think a lot of this code could be refactored by implementing these operations (subsample, convert to single state, make peaky) as flow-nodes instead. This would perhaps be more elegant from a design perspective and also give the user more control.

ZhouW321:
theoretically yes. This has to be done before alignment is input to the Extractor/Processor. Then it would require one/several flow nodes with the same full capability as the Extractor/Processor w.r.t. AM, allophone and state-tying, etc., to parse and modify the alignment. This will be basically copying the same code/members of Extractor/Processor to that flow node class and copy over the same modify function (subsample, make peaky, etc.) there as well. And it requires the users to additionally setup the flow files instead of just one config parameters, which IMO would even increase difficulty and need additional effort to setup workflow manager, e.g. sisyphus.

SimBe195:
Ah, that's true, state-tying has to be applied before. Yeah, that would make it way too complicated.

src/Nn/BufferedAlignedFeatureProcessor.cc
```diff
@@ -159,18 +196,188 @@ void BufferedAlignedFeatureProcessor<T>::generateMiniBatch(std::vector<NnMatrix>
     if (weightedAlignment_)
         miniBatchAlignmentWeights.resize(batchSize, 0);
 
+    u32 idx = 0;
     for (u32 i = 0; i < batchSize; i++) {
```
SimBe195, line 200: Variable names `i` and `idx` next to each other are not so nice.

src/Nn/BufferedAlignedFeatureProcessor.cc
```diff
@@ -159,18 +196,188 @@ void BufferedAlignedFeatureProcessor<T>::generateMiniBatch(std::vector<NnMatrix>
...
+        {   // might have batch offset
+            u32 label = alignmentBuffer_.at(PrecursorBuffer::nProcessedFeatures_ + i);
+            if ( label != lastLabel )
+            {   // supress silence to reduce block conflict ?
+                //if ( !labelIdx.empty() )
+                //{ if ( label == silId )
+                //    continue;
+                //  if ( lastLabel == silId )
+                //    labelIdx.pop_back();
+                //}
```
SimBe195, line 241: Commented code can be removed?

src/Speech/AligningFeatureExtractor.cc
```diff
@@ -164,14 +187,20 @@ bool AligningFeatureExtractor::initializeAlignment() {
         error("Failed to extract alignment.");
         return false;
     }
+    if ( peakyAlignment_ )
+        makePeakyAlignment(&alignmentRef_->data());
+    else if ( forceSingleState_ )
+        makeSingleStateAlignment(&alignmentRef_->data());
```
SimBe195, line 193: How "peaky-alignment" and "force-single-state" play together is not obvious without looking at the code. At the moment "peaky-alignment" overwrites "force-single-state" in that it always also reduces the alignment to single-state regardless of the value of `forceSingleState_`. I think it would be clearer if "peaky-alignment" reduces the alignment to single-state if and only if "force-single-state" is set to true, then they would have their independent effects. This would also allow using peakiness for 3-state alignments in an obvious way. The `makePeakyAlignment` function can be easily extended to preserve the states, so I would change it to a template function like
`template <bool forceSingleState> void AligningFeatureExtractor::makePeakyAlignment` and then change this case distinction to
```suggestion
    if ( peakyAlignment_ ) {
        if ( forceSingleState_ )
            makePeakyAlignment<true>(&alignmentRef_->date());
        else
            makePeakyAlignment<false>(&alignmentRef_->data());
    } else if ( forceSingleState_ )
        makeSingleStateAlignment(&alignmentRef_->data());
```

SimBe195 (changes requested):

src/Speech/AligningFeatureExtractor.cc
ZhouW321: yes, peaky alignment is not necessarily single state and these can be separated.  As  forceSingleState_ is a data member already and there's not much overhead with the for loop over alignments, a template function is not necessary here.

ZhouW321:
no need anymore

PR #27, opened 2022-07-13
Remove TreeSearch unpushedCoarticulatedRootStates assertion

Atticus1806:
When running a `AdvancedTreeSearchLmImageAndGlobalCacheJob` in a basic pipeline (cf. https://github.com/rwth-i6/i6_experiments/pull/74/) there is the following assertion error:

```cpp
PROGRAM DEFECTIVE:
assertion !network.unpushedCoarticulatedRootStates.empty() violated
in void Search::StaticSearchAutomaton::buildBatches() file SearchSpace.cc line 729
```

after speaking with @christophmluscher @Marvin84 and @SimBe195 I got told that this issue is known and commenting this out is a quick fix. Since other people might encouter this I am opening this PR as a draft so they can use the fix, but I am aware that this most likely is not the intended (full) fix.

### DIFF
--- a/src/Search/AdvancedTreeSearch/SearchSpace.cc
+++ b/src/Search/AdvancedTreeSearch/SearchSpace.cc
@@ -726,7 +726,7 @@ void StaticSearchAutomaton::buildBatches() {
         network.dumpDotGraph(paramDumpDotGraph(config), stateDepths);
 
     // Print some useful statistics about pushed and unpushed labels
-    verify(!network.unpushedCoarticulatedRootStates.empty());
+    // verify(!network.unpushedCoarticulatedRootStates.empty());
 
     u32 unpushedLabels = 0;
     u32 pushedLabels   = 0;
--- a/src/Search/AdvancedTreeSearch/SearchSpace.cc
+++ b/src/Search/AdvancedTreeSearch/SearchSpace.cc
@@ -726,6 +726,7 @@ void StaticSearchAutomaton::buildBatches() {
         network.dumpDotGraph(paramDumpDotGraph(config), stateDepths);
 
     // Print some useful statistics about pushed and unpushed labels
+    // this assertion is not needed and leads to crashes. not clear why
     // verify(!network.unpushedCoarticulatedRootStates.empty());
 
     u32 unpushedLabels = 0;


src/Search/AdvancedTreeSearch/SearchSpace.cc
```diff
@@ -726,7 +726,7 @@ void StaticSearchAutomaton::buildBatches() {
         network.dumpDotGraph(paramDumpDotGraph(config), stateDepths);
 
     // Print some useful statistics about pushed and unpushed labels
-    verify(!network.unpushedCoarticulatedRootStates.empty());
+    // verify(!network.unpushedCoarticulatedRootStates.empty());
```
christophmluscher, line 729:
```suggestion
    // this assertion is not needed and leads to crashes. not clear why
    // verify(!network.unpushedCoarticulatedRootStates.empty());
```

christophmluscher (approved):
if we want to merge as is, I would suggest adding a comment along my suggestion

christophmluscher:
@SimBe195 could you check why this assertion fails? there is also a discussion in `slack #rasr`

JackTemaki:
@curufinwe @Marvin84 @SimBe195 Ping :)

SimBe195:
This issue s not present in AppTek RASR due to some changes that were adapted in by https://github.com/rwth-i6/rasr/pull/32, so this PR is superseded by the other one.

christophmluscher:
Which RASR commit are you using?

JackTemaki:
He is using my RASR for Ubuntu 22, latest master. I had the same problem when running LS960 again using the master only.

SimBe195:
I checked the setup of @Atticus1806 and the RASR that has been used was indeed outdated. With the current master the job is able to complete without issues.

JackTemaki:
Indeed I gave @Atticus1806 the wrong version, but I have the error with a more current one from November.

JackTemaki:
See: `/work/common/asr/librispeech/data/sisyphus_export_setup/alias/baselines/librispeech/ls960/gmm/common_baseline/recognition/train-other-960-mono/recog_train-other-960-mono-dev-clean-ps1.00-lm10.00-iter10`

SimBe195:
@JackTemaki The `AdvancedTreeSearchJob` that you linked uses a more modern RASR but the `AdvancedTreeSearchLmImageAndGlobalCacheJob` that it's based on was run with an older RASR version from July 2022 where the assertion was commented out instead of applying the fix from #32.
Re-running the `AdvancedTreeSearchLmImageAndGlobalCacheJob` with a modern RASR version fixes the issue.

JackTemaki:
> @JackTemaki The `AdvancedTreeSearchJob` that you linked uses a more modern RASR but the `AdvancedTreeSearchLmImageAndGlobalCacheJob` that it's based on was run with an older RASR version from July 2022 where the assertion was commented out instead of applying the fix from #32. Re-running the `AdvancedTreeSearchLmImageAndGlobalCacheJob` with a modern RASR version fixes the issue.

Oh well.... That explains why it only failed for the monophone jobs and not for the rest, I guess I changed it in between then... thanks a lot for figuring this out....

PR #28, opened 2022-08-04
Add remove-corpus-name-prefix parameter

JackTemaki:
This is a draft PR for code that allows to remove the top-level/super corpus name from the segment full-name. 

This can be used to merge bliss files using the subcorpora mode while at the same time just merging feature and alignment bundle files without the necessity to rename all entries and create new files.

I am open to better naming suggestions of the parameter.

~~I am also not sure if I can hijack the "anonymous" default value like that, but this seemed to be the easiest approach so far.~~

### DIFF
--- a/src/Bliss/CorpusDescription.cc
+++ b/src/Bliss/CorpusDescription.cc
@@ -51,13 +51,13 @@ std::string NamedCorpusEntity::fullName() const {
         new_name = parent()->fullName() + "/" + name();
     else
         new_name = name();
-
     if (!removePrefix_.empty()) {
         auto res = std::mismatch(removePrefix_.begin(), removePrefix_.end(), new_name.begin());
         if (res.first == removePrefix_.end()){
             new_name = std::string(res.second, new_name.end());
         }
     }
+    return new_name;
 }
 
 // ========================================================================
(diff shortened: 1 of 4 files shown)


src/Bliss/CorpusDescription.cc
```diff
@@ -42,13 +42,22 @@ const char* const NamedCorpusEntity::anonymous = "ANONYMOUS";
...
     else
-        return name();
+        new_name = name();
+
+    if (!removePrefix_.empty()) {
+        auto res = std::mismatch(removePrefix_.begin(), removePrefix_.end(), new_name.begin());
+        if (res.first == removePrefix_.end()){
+            new_name = std::string(res.second, new_name.end());
+        }
+    }
```
SimBe195, line 60: The return is missing.

JackTemaki:
@SimBe195 sorry I pushed something incomplete last time, now it is tested an working

SimBe195 (approved):

curufinwe:
Why implement it as a prefix to be removed and not as "please do not use the name of the corpus object"? With this parameter you need to make sure to get the corpus name right. If you just have a boolean parameter to remove the corpus name you can't make this mistake.

JackTemaki:
> Why implement it as a prefix to be removed and not as "please do not use the name of the corpus object"? With this parameter you need to make sure to get the corpus name right. If you just have a boolean parameter to remove the corpus name you can't make this mistake.

I implemented this first (see history), but then Simon and Wei argued for the more flexible variant.

SimBe195:
> Why implement it as a prefix to be removed and not as "please do not use the name of the corpus object"? With this parameter you need to make sure to get the corpus name right. If you just have a boolean parameter to remove the corpus name you can't make this mistake.

With the way it is now you can also remove not only the corpus name but also more complex subcorpus hierarchies if you want to.

PR #31, opened 2022-10-21
Fix encoding issues when writing CART examples/trees

curufinwe:
backported from AppTek codebase

### DIFF
--- a/src/Cart/DecisionTree.cc
+++ b/src/Cart/DecisionTree.cc
@@ -193,6 +193,9 @@ void DecisionTree::Path::write(std::ostream& out) const {
 const Core::ParameterString DecisionTree::paramCartFilename(
         "decision-tree-file",
         "name of decision tree (aka cart) file");
+const Core::ParameterString DecisionTree::paramCartEncoding(
+        "encoding",
+        "encoding of the cart file");
 
 DecisionTree::DecisionTree(const Core::Configuration& config,
                            PropertyMapRef             map,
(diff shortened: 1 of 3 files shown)


Marvin84 (approved):

src/Cart/DecisionTree.cc
```diff
@@ -193,6 +193,9 @@ void DecisionTree::Path::write(std::ostream& out) const {
 const Core::ParameterString DecisionTree::paramCartFilename(
         "decision-tree-file",
         "name of decision tree (aka cart) file");
+const Core::ParameterString DecisionTree::paramCartEncoding(
+        "encoding",
+        "encoding of the cart file");
```
albertz, line 198: What is the default? Maybe set utf8 as default?

Marvin84: Without this fix the default encoding turns out to be `ISO-8859-1` and as my current understanding, this is hardcoded under `Core::Unicode`.

albertz: Should we change this default? EIther here, or directly in `Core::Unicode`?

Marvin84: I changed the default to `utf-8` in my local rasr, since I thought this is a more useful default. This also solved the issue for me. But I do not have a good overview of all use cases. Maybe @curufinwe can elaborate on a better answer.

SimBe195: I would be fine to change the default also in `Core::Unicode`, as long as this doesn't break anything.

curufinwe: For sisyphus setups it will probably not break as we set encoding to utf-8 there anyway. But legacy setups / bash-setups might break depending if they have encoding set or not. Thus I would not change the default now as sisyphus setups consistently set encoding anyway.

SimBe195: Okay, then we will just keep this as it is for now.

SimBe195 (approved):

PR #38, opened 2022-11-17
Dense tying for mono/di/tri-phone with end of word classes

Marvin84:
This completes different dense tyings and adds the possibility of using the end of word classes for the dense tying. This type of tying is used mostly for the factored hybrid model.

1. I corrected the already implemented `DiphoneNoTyingDense` name into `DiphoneDense`. I assume that the whole set of triphone state classes is equivalent to no tying, as soon as we reduce the context to first or zero order, for diphone and monophone, respectively, it might be better to avoid appending `Notying` for better understandability.
2. `MonophoneDense` tying has two additional labels for silence, and three additional labels corresponding to `#` which are effectively never used. This is important since keeps the mapping consistent when one switches to a higher phonetic order.


src/Am/ClassicStateTying.cc
```diff
@@ -181,8 +181,15 @@ bool LutStateTying::loadLut(const std::string& filename) {
 // ============================================================================
 
 const u32 NoStateTyingDense::numBoundaryClasses_ = 4;  // 0: none, 1: start (@i), 2: end (@f), 3: start+end (@i@f)
+const u32 NoStateTyingDense::numWordEndClasses_ = 2; // 0: {none, start (@i)}, 1: {end (@f), 3: start+end (@i@f)}
 
-const Core::ParameterBool NoStateTyingDense::paramUseBoundaryClasses("use-boundary-classes", "wether boundary classes should be considered during state-tying", true);
+const Core::ParameterBool NoStateTyingDense::paramUseBoundaryClasses("use-boundary-classes",
+                                                                     "wether boundary classes should be considered during state-tying",
```
SimBe195, line 187: There is a typo in "whether". Maybe you can fix this while you're at it.

src/Am/ClassicStateTying.cc
```diff
@@ -236,6 +248,13 @@ Mm::MixtureIndex NoStateTyingDense::classify(const AllophoneState& a) const {
         result *= numBoundaryClasses_;
         result += a.allophone()->boundary;
     }
+    if (useWordEndClasses_) {
+        result *= numWordEndClasses_;
+        u32 weClass = !(((a.allophone()->boundary == Am::Allophone::isInitialPhone) or
+                         (a.allophone()->boundary == Am::Allophone::isWithinPhone)) &&
+                        (a.allophone()->boundary != Am::Allophone::isFinalPhone));
```
SimBe195, line 255: Can't this be simplified to just
`u32 weClass = a.allophone()->boundary & Am::Allophone::isFinalPhone;`?

SimBe195 (changes requested):

src/Am/ClassicStateTying.cc
```diff
@@ -285,9 +307,66 @@ Mm::MixtureIndex DiphoneNoStateTyingDense::classify(const AllophoneState& a) con
...
+    }
     require_lt(result, nClasses_);
 
     return result;
 }
 
-}  // namespace Am
+// ============================================================================
+
+MonophoneDense::MonophoneNoStateTyingDense(const Core::Configuration& config, ClassicStateModelRef stateModel)
```
SimBe195, line 324: This should be `MonophoneDense::MonophoneDense`.

src/Am/ClassicStateTying.cc
```diff
@@ -181,8 +181,15 @@ bool LutStateTying::loadLut(const std::string& filename) {
...
+const u32 NoStateTyingDense::numWordEndClasses_ = 2; // 0: {none, start (@i)}, 1: {end (@f), 3: start+end (@i@f)}
 
-const Core::ParameterBool NoStateTyingDense::paramUseBoundaryClasses("use-boundary-classes", "wether boundary classes should be considered during state-tying", true);
+const Core::ParameterBool NoStateTyingDense::paramUseBoundaryClasses("use-boundary-classes",
+                                                                     "wether boundary classes should be considered during state-tying",
+                                                                     true);
+
+const Core::ParameterBool NoStateTyingDense::paramUseWordEndClasses("use-word-end-classes",
+                                                                    "whether word end class should be considered during state-tying",
+                                                                    false);
```
SimBe195, line 192: This parameter is still missing in the header file.

src/Am/ClassicAcousticModel.cc
```diff
@@ -106,7 +106,8 @@ const Core::Choice ClassicAcousticModel::choiceStateTying(
         "none", noTying,
         "no-tying", noTying,
         "no-tying-dense", noTyingDense,
-        "diphone-no-tying-dense", diphoneNoStateTyingDense,
+        "diphone-dense", diphoneDense,
```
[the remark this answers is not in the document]
Marvin84: We had just one HiWi using this and he did not publish any recipe, no official setup is going to break.

src/Am/ClassicStateTying.cc
Marvin84: I do not understand your proposed solution. An allophone can have different boundary classes and the second argument is one possible class (btw, with a value equal to 2), how would this decide whether you have an allophone of word end class or not?

SimBe195: As I understand, the boundary is represented as a bit-mask. The 4 options are:
`"" = 00b = 0 = isWIthinPhone`
`"@i" = 01b = 1 = isInitialPhone`
`"@f" = 10b = 2 = isFinalPhone`
`"@i@f" = 11b = 3 = isInitialPhone | isFinalPhone"`
So using bit-wise 'and' with `isFinalPhone` you can find out if the boundary is an EOW class, i. e. option 2 or 3.

Marvin84: But this does not give me 0 or 1 as result

SimBe195: Yeah that's true, you need to cast to boolean as an in-between step. I think the simplest would be
`u32 weClass = !!(a.allophone()->boundary & Am::Allophone::isFinalPhone);`.

Marvin84: But then this is also a two-step operation, not so much simpler. If you agree, I would just stick to what is here which is probably more readable for a user who does not know how each boundary class is implemented. What is your take?

src/Am/ClassicStateTying.cc
```diff
@@ -191,11 +198,16 @@ NoStateTyingDense::NoStateTyingDense(const Core::Configuration& config, ClassicS
           numStates_(stateModel->getHmmTopologySet()->getDefault().nPhoneStates()),                                                 // usually 3
           contextLength_(std::max(stateModel->phonology().maximumHistoryLength(), stateModel->phonology().maximumFutureLength())),  // usually 1
           useBoundaryClasses_(paramUseBoundaryClasses(config)),
+          useWordEndClasses_(paramUseWordEndClasses(config)),
           nClasses_(0) {
+    require(!(useBoundaryClasses_ &&  useWordEndClasses_));
```
NeoLegends, line 203: Nit:
```suggestion
    require(!(useBoundaryClasses_ && useWordEndClasses_));
```

src/Am/ClassicStateTying.cc
```diff
@@ -276,18 +297,72 @@ Mm::MixtureIndex DiphoneNoStateTyingDense::classify(const AllophoneState& a) con
...
+
+    return result;
+}
+
+// ============================================================================
+
+MonophoneDense::MonophoneDense(const Core::Configuration& config, ClassicStateModelRef stateModel)
+        : Core::Component(config),
+          NoStateTyingDense(config, stateModel) {
+    require(!(useBoundaryClasses_ &&  useWordEndClasses_));
```
NeoLegends, line 325:
```suggestion
    require(!(useBoundaryClasses_ && useWordEndClasses_));
```

NeoLegends (commented):
The state tying impl looks good to me, I just left very minor nits.

Maybe this is not the right place to discuss this, but generally I wonder if we can improve on the `useBoundaryClasses` / `useWordEndClasses` approach. Currently we use two separate booleans, so we have four possible combinations of `true/false`, but only three valid combinations. This sounds more like a use case for an enum with three states to me.

Maybe this ship has already sailed because we'd need to adapt so much code elsewhere, but what would you think about a `use-word-end-boundary-classes` config parameter that can take the values `none` / `word-end` / `boundary` instead of the two boolean flags (in a follow up PR maybe)?

NeoLegends (approved):

Marvin84:
> This sounds more like a use case for an enum with three states to me.

Good point, I agree. The only problem is that there might be legacy setups that will break due to hash changes on sisyphus side. However, since still few people are using the dense tying, we might still be on time to do this change. Let's see what @curufinwe and @SimBe195 think about this.

NeoLegends:
> [ADD] I can anyhow guarantee one rasr version under a separate branch that has still the old behavior for such cases.

Perhaps this is a case where we should make RASR backwards compatible, so that it supports both ways of setting the WE/BD class configuration (in one branch, seems easy enough). Internally RASR would use the three-state enum, but externally (in the CRP) it can be configured both ways as long as the user choses one variant or the other.

This way we can keep the hashes of existing setups and we're less error prone inside of RASR.

SimBe195:
If, like you said, (almost) no one uses these tyings so far I would be fine with losing the backwards compatibility here. The renaming of `"diphone-no-tying-dense"` is anyway also a change that breaks old setups. Ideal would be either a 3- (or 4-)-state enum or two boolean flags like `"use-word-end-classes"` and `"use-word-begin-classes"` which are not mutually exclusive. If both of these flags are true this would be equivalent to the `"use-boundary-classes"` flag that we have now.

src/Am/ClassicStateTying.cc
```diff
@@ -276,18 +297,72 @@ Mm::MixtureIndex DiphoneNoStateTyingDense::classify(const AllophoneState& a) con
     result *= numStates_;
     result += u32(a.state());
 
+    if (useBoundaryClasses_) {
+        result *= numBoundaryClasses_;
+        result += a.allophone()->boundary;
+    }
+    if (useWordEndClasses_) {
+        result *= numWordEndClasses_;
+        u32 weClass = a.allophone()->boundary & Am::Allophone::isFinalPhone ? 1 : 0;
```
curufinwe, line 306: I would add brackets around the check. s.t. it's unambiguous (for those who don't know operator precedence by heart) that the `&` is evaluated before the `?`

src/Am/ClassicStateTying.cc
```diff
@@ -276,18 +297,72 @@ Mm::MixtureIndex DiphoneNoStateTyingDense::classify(const AllophoneState& a) con
...
+    result *= numStates_;
+    result += u32(a.state());
+
     if (useBoundaryClasses_) {
         result *= numBoundaryClasses_;
         result += a.allophone()->boundary;
     }
+    if (useWordEndClasses_) {
+        result *= numWordEndClasses_;
+        u32 weClass = a.allophone()->boundary & Am::Allophone::isFinalPhone ? 1 : 0;
```
curufinwe, line 361: see above

curufinwe (changes requested):
Only one minor thing.

curufinwe (approved):

PR #39, opened 2022-11-28
flexible fullsum training automaton: HMM/CTC/RNA topology + modifications

ZhouW321:
provide various configurable automaton to RETURNN for fullsum training


ZhouW321:
do it in Speech/AllophoneStateGraphBuilder  so that both training and alignment use the same code to get automaton

PR #45, opened 2023-02-22
Generic Seq2Seq Decoder

ZhouW321:
might still run auto clang formatting

PR #49, opened 2023-03-29
Generic Seq2Seq Decoder preparation

SimBe195:
This PR contains a subset of changes from #45. These are all the smaller file changes that prepare the rest of the code base for the new module but omitting actual implementations of the new label scorers and search/alignment algorithms.

The code in this PR can be compiled successfully but the new classes are mostly empty and non-functional at this stage.


src/Lattice/Lattice.cc
```diff
@@ -239,6 +248,18 @@ void StandardWordLattice::newArc(
     lm_->state(source->id())->newArc(target->id(), Fsa::Weight(lm), (lemmaPronunciation) ? lemmaPronunciation->id() : Fsa::Epsilon);
 }
 
+void StandardWordLattice::newArc(
```
curufinwe, line 251: We should consider refactoring this to use a new private method that takes an index instead of of a pointer to Lemma/LemmaPronunciation and call it in the public newArc functions.

curufinwe (changes requested):

src/Speech/ModelCombination.hh
```diff
@@ -37,10 +39,11 @@ namespace Speech {
 class ModelCombination : public Mc::Component, public Core::ReferenceCounted {
 public:
     typedef u32       Mode;
-    static const Mode complete;
+    static const Mode complete;  // Includes lexicon, AM and LM but NOT label scorer; named 'complete' for legacy reasons.
```
SimBe195, line 42: While the name `complete` is a bit misleading now because it doesn't include the label scorer, changing it would mean a config-level change since the name 'complete' is also used in `src/Speech/LatticeNodes.cc` for a `Core::Choice ModelCombinationNode::choiceMode`. So a comment was added here and the name is left as-is for now.

PR #50, opened 2023-04-14
Connect word ends to initial state for context-independent allophones

SimBe195:
For context-independent allophones (i.e. history_length = future_length = 0) the word-end states were not connected to the initial state in the `ClassicTransducerBuilder` which results in a broken automaton and a crash during training with `fast_bw` loss (e. g. CTC).
A fix for this behaviour had already been implemented in the original `label_sync_decoding` branch of @ZhouW321 but was missed when porting to github RASR. This PR re-applies the original fix.

### DIFF
--- a/src/Am/ClassicTransducerBuilder.cc
+++ b/src/Am/ClassicTransducerBuilder.cc
@@ -155,10 +155,12 @@ ClassicTransducerBuilder::ClassicTransducerBuilder(Core::Ref<
     fixAllophoneContextAtWordBoundaries_    = paramFixAllophoneContextAtWordBoundaries(model_->getConfiguration());
     statistics_                             = new Statistics;
 
-    if ( model_->phonology()->maximumHistoryLength() == 0 && model_->phonology()->maximumFutureLength() == 0 )
-      linkWordEndStart_ = true;
-    else
-      linkWordEndStart_ = false;
+    if (model_->phonology()->maximumHistoryLength() == 0 && model_->phonology()->maximumFutureLength() == 0) {
+        linkWordEndStart_ = true;
+    }
+    else {
+        linkWordEndStart_ = false;
+    }
 }
 
 ClassicTransducerBuilder::~ClassicTransducerBuilder() {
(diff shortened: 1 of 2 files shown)


mmz33:
Hey @curufinwe, we need this urgently for some experiments so could you please review this soon if possible? thank you.

src/Am/ClassicTransducerBuilder.cc
```diff
@@ -154,6 +154,11 @@ ClassicTransducerBuilder::ClassicTransducerBuilder(Core::Ref<
     filterOutInvalidAllophones_             = paramFilterOutInvalidAllophones(model->getConfiguration());
     fixAllophoneContextAtWordBoundaries_    = paramFixAllophoneContextAtWordBoundaries(model_->getConfiguration());
     statistics_                             = new Statistics;
+
+    if ( model_->phonology()->maximumHistoryLength() == 0 && model_->phonology()->maximumFutureLength() == 0 )
+      linkWordEndStart_ = true;
+    else
+      linkWordEndStart_ = false;
```
curufinwe, line 161:
```suggestion
    if (model_->phonology()->maximumHistoryLength() == 0 && model_->phonology()->maximumFutureLength() == 0) {
        linkWordEndStart_ = true;
    }
    else {
        linkWordEndStart_ = false;
    }
```

curufinwe (commented):
I'm currently not super familiar with this part of the code. Could you attach plots of a simple automaton before and after your fix? Thanks!

Marvin84:
Yes, me and @michelwi are aware of this problem, since the FSA that uses allophones without right and left history is broken. I agree that there should be more documentation and test that confirms the correctness of the resulting FSA and I also suggest that this PR includes `SVG`s of an example. 

I personally find that this solution is hacky and mathematically incorrect. The classic FST creation is constructing an `HCLG` graph, and the developer of seq2seq framework needs an `HLG` graph, but by making all right and left context to be equal to `#` you are not creating an `HLG` graph, you are simply hacking the system for a naive and simplified solution.

albertz:
Note that this fix here is needed to fix https://github.com/rwth-i6/rasr/issues/68 and https://github.com/rwth-i6/returnn/issues/1456.

Marvin84:
Regarding the `allophoneStateToPhonemeTransducer ` I am attaching here the resulting FSA from our standard HMM with the proper composition. As you can see, there are no dead-end nodes, meaning all paths will finish in one final node. I believe this is one first trivial condition for a proper FSA definition. The question is how this current composition affects the optimization and convergence when for example doing full-sum training. 

allophon

curufinwe (approved):

PR #52, opened 2023-04-25
Allow 24-bit acoustic model indices

NeoLegends:
This PR allows for 24-bit acoustic model indices/state indices in the same 32-bit as previously used for the `StateDesc` struct by using bit fields to make use of the padding in the struct. In particular, this now allows for triphone factored hybrid/no-tying-dense labels.

2^24 = 16,777,216.

Here is an example (Rust) playground, which demonstrates the size of the struct stays the same: https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=dc00f470e0cfa3272d02b00f96a04e59. The example uses the same layout rules as C/C++ would.

Credits: I think these changes here originally come from Eugen. I was only made aware that `StateDesc` size is a limitation a short while ago and I'm merely implementing the changes here.


curufinwe (changes requested):
1. Have you tested if there is a performance impact from this change?
2. Please at least update the DiskFormatVersion inside TreeStructure.cc, better update the code to also load the previous version. In any case have a test-setup where you make sure that RASR behaves well with an old global-cache.

Marvin84:
@NeoLegends  I agree with @curufinwe that this needs to be first tested. I add also @SimBe195 for his input.
My understanding is that there are different function calls that now need to read properly the acoustic and transition indices and the mere change of the struct as in this PR might not be enough.

src/Search/StateTree.cc
```diff
@@ -882,9 +882,9 @@ StateTree::State::State() {
     successors = invalidBatch;
 }
 
-const StateTree::Depth                         StateTree::invalidDepth         = Core::Type<StateTree::Depth>::max;
-const StateTree::BatchId                       StateTree::invalidBatch         = Core::Type<StateTree::BatchId>::max;
-const StateTree::StateDesc::AcousticModelIndex StateTree::invalidAcousticModel = Core::Type<StateTree::StateDesc::AcousticModelIndex>::max;
+const StateTree::Depth                 StateTree::invalidDepth         = Core::Type<StateTree::Depth>::max;
+const StateTree::BatchId               StateTree::invalidBatch         = Core::Type<StateTree::BatchId>::max;
+const StateTree::StateDesc::ModelIndex StateTree::invalidAcousticModel = Core::Type<StateTree::StateDesc::ModelIndex>::max;
```
curufinwe, line 887: I'm not sure if this is correct here. Some states might store invalidAcousticModel in the state-tree and the value here is 2^32 - 1, but the maximum value that can be read from StateDesc::acousticModel is 2^24 -  1.

src/Search/StateTreeIo.cc
```diff
@@ -36,14 +36,26 @@ void StateTreeIo::write<StateTree::Exit>(Output& o, const StateTree::Exit& exit)
 
 template<>
 void StateTreeIo::read<StateTree::StateDesc>(Input& i, StateTree::StateDesc& desc) const {
-    read(i, desc.acousticModel);
-    read(i, desc.transitionModelIndex);
+    StateTree::StateDesc::ModelIndex           am = 0;
+    StateTree::StateDesc::TransitionModelIndex tm = 0;
+
+    read(i, am);
```
curufinwe, line 42: I think it's better to store the data as one 32-bit value and then write the upper 8 bit into transitionModel index and the lower 24 bits into acousticModel

src/Search/StateTreeIo.cc
```diff
@@ -36,14 +36,26 @@ void StateTreeIo::write<StateTree::Exit>(Output& o, const StateTree::Exit& exit)
...
 }
 
 template<>
 void StateTreeIo::write<StateTree::StateDesc>(Output& o, const StateTree::StateDesc& desc) const {
-    write(o, desc.acousticModel);
-    write(o, desc.transitionModelIndex);
+    StateTree::StateDesc::ModelIndex           am = desc.acousticModel;
+    StateTree::StateDesc::TransitionModelIndex tm = desc.transitionModelIndex;
+
+    write(o, am);
```
curufinwe, line 57: This also needs to be changed then (see above).

src/Search/StateTreeIo.cc
```diff
@@ -77,7 +89,7 @@ void StateTreeIo::write<StateTree::CoarticulationStructure::PhonemePair>(
 }
 
 const std::string StateTreeIo::magic             = "SPRINT-ST";
-const int         StateTreeIo::fileFormatVersion = 5;
+const int         StateTreeIo::fileFormatVersion = 6;
```
curufinwe, line 92: Is this checked anywhere? If so adding backwards compatibility would be nice.

curufinwe (changes requested):

src/Search/StateTree.cc
```diff
@@ -921,7 +921,7 @@ StateTree::StateTree(const Core::Configuration&         c,
         earlyRecombinationLimit_ = paramEarlyRecombinationLimit(config);
     }
 
-    if (acousticModel_->nStateTransitions() >= Core::Type<StateDesc::AcousticModelIndex>::max) {
+    if (acousticModel_->nStateTransitions() >= StateDesc::MAX_AM) {
```
curufinwe, line 924: This is incorrect, the maximum number of transition models is much lower than the maximum number of acoustic models. This should be `Core::Type<StateDesc::TransitionModelIndex>::max`

curufinwe:
I have tested your implementation by merging it into the AppTek-rasr codebase and running some of our models with it. I fixed the bugs that I found and think that now it's ready to be merged.

curufinwe (approved):

NeoLegends:
🎉

PR #54, opened 2023-05-08
Apptainer specifications tf2.8

SimBe195:
This PR adds specifications for an apptainer image that can be used to compile and use RASR with tensorflow. Included are:

- An apptainer definition file `image.def`
- A `makefiles.patch`file that can be applied by git in order to correctly set library paths, enable/disable modules etc. for the image in the RASR makefiles.
- A `README.txt` file that describes the image produced by the `image.def` file.

The directory structure is made in a way that makes it easy to add new image specifications in the future.
Also, outdated pre-existing singularity `.def` files were removed.


JackTemaki (commented):
Nice :) I have some additional questions though:
 - If you anyway re-build Tensorflow, would it be possible to include MKL?
 - Would it make sense to add a `./configure` flag, e.g. `--patchfile` or so to automatically update the build config?
 - What about the FFMPEG node, did you have problems with the libraries?

SimBe195:
> If you anyway re-build Tensorflow, would it be possible to include MKL?
Maybe it's possible but it's not trivial. I have tried by simply adding "--config=mkl" to the bazel build but that resulted in API mismatches; some error like `error: no matching function for call to ... note:   candidate expects 9 arguments, 8 provided` when compiling MKL ops. Maybe there's a reason that MKL was not already included in the official NVidia image...

> Would it make sense to add a `./configure` flag, e.g. `--patchfile` or so to automatically update the build config?
Sure, I can do that!

> What about the FFMPEG node, did you have problems with the libraries?
Yes, the ffmpeg module in RASR was built with an outdated version of the API and updating to a newer ffmpeg version requires some changes in the RASR source code in `src/Audio/Ffmpeg.cc`.

apptainer/2023-05-08_tensorflow-2.8_v1/makefiles.patch
```diff
@@ -0,0 +1,118 @@
...
+ LDFLAGS     += -Wl,-rpath -Wl,${PYTHON_PATH}/lib
+ else
+ INCLUDES    += `python3-config --includes 2>/dev/null`
+-LDFLAGS     += `python3-config --ldflags 2>/dev/null`
++LDFLAGS     += `python3-config --ldflags --embed 2>/dev/null`
+ # IF you want to use Python2 for whatever reason:
+ # INCLUDES    += `pkg-config --cflags python`
+ # LDFLAGS     += `pkg-config --libs python`
+ endif
++INCLUDES    += -I/usr/local/lib/python3.8/dist-packages/numpy/core/include
```
JackTemaki, line 115: we could also use:
`INCLUDES    += -I$(shell python3 -c 'import numpy as np; print(np.get_include())')` in the general config, this should always be correct given the right python binary is sourced.

apptainer/2023-05-08_tensorflow-2.8_v1/makefiles.patch
```diff
@@ -0,0 +1,118 @@
...
+diff --git a/config/cc-gcc.make b/config/cc-gcc.make
+index bb6b97e..8f1ae43 100644
+--- a/config/cc-gcc.make
++++ b/config/cc-gcc.make
+@@ -26,7 +26,7 @@ CCFLAGS		+= -pipe
+ CCFLAGS		+= -funsigned-char
+ CCFLAGS		+= -fno-exceptions
+ CFLAGS		+= -std=c99
+-CXXFLAGS	+= -std=gnu++0x
++CXXFLAGS	+= -std=c++14
```
JackTemaki, line 46: I think can be safely changed in the main config or not?

apptainer/2023-05-08_tensorflow-2.8_v1/makefiles.patch
SimBe195: I think the shell command is fine. Implemented the suggestion.

SimBe195:
>     * Would it make sense to add a `./configure` flag, e.g. `--patchfile` or so to automatically update the build config?

I modified the configure script; now it can be called like
`./configure --apptainer-patch=2023-05-08_tensorflow-2.8_v1`
to apply the patch file.

SimBe195:
> * don't delete old singulariy definition files + add fixes for c++ version
> 
> * make patches full replacements -> via generic shell script

Suggestions have been applied. Old singularity definitions are now available under `apptainer/2022-10-21_tensorflow-1.15_v1` and `apptainer/2022-10-21_tensorflow-1.15_arm_v1` respectively again and makefiles are now set-up using the `setup_makefiles.sh` script in each apptainer subdirectory instead of a patch file. At the moment all three `setup_makefiles.sh` link a new script under `scripts/setup_apptainer_makefiles.sh` which simply copies over all make-related files from the respective `makefiles` subdirectory to the project base.

apptainer/2023-05-08_tensorflow-2.8_v1/image.def
```diff
@@ -0,0 +1,32 @@
...
+    sed -i "s/D_GLIBCXX_USE_CXX11_ABI=0/D_GLIBCXX_USE_CXX11_ABI=1/" nvbuildopts  # Enable CXX11 ABI
+    ./nvbuild.sh --v2 --sm 6.1,7.5,8.6  # GTX 1080, RTX 2080, RTX 3090
+
+    ln -s /usr/local/lib/tensorflow/libtensorflow_cc.so.2 /usr/local/lib/tensorflow/libtensorflow_cc.so
+    ln -s /usr/local/lib/python3.8/dist-packages/tensorflow/libtensorflow_framework.so.2 /usr/local/lib/tensorflow/libtensorflow_framework.so.2
+    ln -s /usr/local/lib/tensorflow/libtensorflow_framework.so.2 /usr/local/lib/tensorflow/libtensorflow_framework.so
+
+%environment
+    # Location of libtensorflow_cc.so and libtensorflow_framework.so
+    export LD_LIBRARY_PATH=/usr/local/lib/tensorflow/:$LD_LIBRARY_PATH
```
curufinwe, line 32: If you're building your own OS image it's better to add this path into a file e.g. `/etc/ld.so.conf.d/tensorflow.conf` and run ldconfig in the image. This way your setup will not break if the user overwrites LD_LIBRARY_PATH instead of appending to it.

curufinwe (commented):
I have 2 suggested changes:
1. make bazel not eat all the resources: For TF bazels estimates on how much memory a compilation job will use are wrong and at least on my machine it causes problems compiling bazel. And on other machines people are maybe also running things like a web-browser and zoom + slack etc. that all also consume quite a bit of memory and might run into this issue too
2. I think it's better to add the tensorflow libraries to the system-wide configuration for shared objects instead of relying on LD_LIBRARY_PATH (one can still overwrite it with LD_LIBRARY_PATH if needed)

curufinwe (approved):

JackTemaki (approved):
The order of flags for configure might be relevant now (you have to do e.g. `--enable-module` after setting apptainer), but this should be okay...

PR #55, opened 2023-05-15
Update to Tensorflow 2.12.0 and Ubuntu 22.04 and C++17

sleepyeldrazi:
Version updates in the singularity containers:
- Ubuntu to 22.04
- Tensorflow to 2.12.0
- Cuda to 11.8
- Cudnn to 8.6

Updates in the config files:
- C++ updated to C++17. `src/Mm/FeatureScorer.hh` updated accordingly.
- Include and library directories for python 3 updated


curufinwe:
No longer needed as we now have a TF 2.14 apptainer image.

PR #56, opened 2023-06-02
Added Tensorflow Lite support

sleepyeldrazi:
Adds Tensorflow Lite support. The version updates from #55 are also included, as Tensorflow 2 is needed for Tensorflow Lite.


curufinwe (changes requested):
* clang-format
* single line blocks (if / while /etc.) should also have curly braces
* private member functions should not start with an underscore

PR #58, opened 2023-06-12
Allow returning subsampled alignments in PythonControl

NeoLegends:
Currently `PythonControl` requires the number of targets it gives to RETURNN matches the number of features. This is a safety assertion to catch data errors. However, it is wrong when training subsampling models that use different time steps for the input features and output labels.

The PR introduces a parameter that sets the subsampling ratio (e.g. 3x, 4x, etc.) and adapts the length validation assertion to respect the subsampling factor accordingly.

### DIFF
--- a/src/Nn/PythonControl.cc
+++ b/src/Nn/PythonControl.cc
@@ -77,6 +77,14 @@ static const Core::ParameterString paramPyModConfig(
         "config-string, passed to init()",
         "");
 
+static const Core::ParameterInt paramSubsamplingFactor(
+        "subsampling-factor",
+        "the subsampling factor for allowing less targets than features",
+        1,
+        1,
+        Core::Type<s8>::max,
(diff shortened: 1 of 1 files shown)


Marvin84 (approved):

SimBe195:
Maybe the assertion can also simply be removed altogether. Training would crash later anyway if the time axis of logits and targets don't match up. I would be fine with getting rid of such additional checks.

NeoLegends:
I don't mind either way.

I agree that RETURNN will also catch issues in the data lengths and that the assertion here is more of a defense-in-depth than really necessary.

I would like not to spend too much time on reviewing this PR, so if you prefer I'll completely remove the assertion.

Marvin84:
I suggest we remove it.

Marvin84 (approved):

SimBe195 (approved):

PR #60, opened 2023-08-09
Add support for Onnx models

curufinwe:


Atticus1806:
Tested with an apptainer image without tensorflow (since the given apptainer would not compile for me) and works with onnxruntime 1.15.1

Atticus1806 (approved):

JackTemaki (approved):

PR #64, opened 2023-08-29
Apptainer tf2_14 image

Atticus1806:
Adds an image with tf 2.13
Tested with manual compilation as well sisyphus based compilation.


curufinwe:
FYI: our tensorflow conformer models don't work with TF 2.13.  The fix was only merged after 2.13 was already out. For AppTek I manually patched TF before compilation to contain this fix. Otherwise looks OK.

Atticus1806:
For now this was for Tedlium Hybrid Baseline in TF, but if you want we can wait for Tf 2.14 and then build a similar image to this. That should contain the fix right?

curufinwe:
You could also add this patch to the tensorflow source code before compiling:
```
diff --git a/tensorflow/core/grappler/optimizers/remapper.cc b/tensorflow/core/grappler/optimizers/remapper.cc    
index 6706f1c97ff..9a8ffe9125d 100644    
--- a/tensorflow/core/grappler/optimizers/remapper.cc    
+++ b/tensorflow/core/grappler/optimizers/remapper.cc    
@@ -1860,6 +1860,18 @@ bool FindSigmoidAndMul(RemapperContext* ctx, int node_index,    
       sigmoidmul_pattern, {}, ctx->graph_view.GetNode(node_index),    
       matched_nodes_map, remove_node_indices);    
     
+  if (found_op_type_match) {    
+    NodeDef* matched_sigmoid_node =    
+        ctx->graph_view.GetNode(matched_nodes_map->at("sigmoid"))->node();    
+    auto in_tensor_sigmoid = matched_sigmoid_node->input(0);    
+    if ((mul_node_def->input(0) != in_tensor_sigmoid) &&    
+        (mul_node_def->input(1) != in_tensor_sigmoid)) {    
+      // If the input tensor of Sigmoid doesn't match with either of input    
+      // tensors of mul return false    
+      found_op_type_match = false;    
+    }    
+  }    
+    
   return found_op_type_match;    
 }    
     
@@ -4185,6 +4197,18 @@ bool FindSoftplusAndTanhAndMul(RemapperContext* ctx, int node_index,    
       softplustanhmul_pattern, {}, ctx->graph_view.GetNode(node_index),    
       matched_nodes_map, remove_node_indices);    
     
+  if (found_op_type_match) {    
+    NodeDef* matched_softplus_node =    
+        ctx->graph_view.GetNode(matched_nodes_map->at("softplus"))->node();    
+    auto in_tensor_softplus = matched_softplus_node->input(0);    
+    if ((mul_node_def->input(0) != in_tensor_softplus) &&    
+        (mul_node_def->input(1) != in_tensor_softplus)) {    
+      // If the input tensor of Softplus doesn't match with either of input    
+      // tensors of mul return false    
+      found_op_type_match = false;    
+    }    
+  }    
+    
   return found_op_type_match;    
 } 
```

Atticus1806:
This image does not compile TF itself, but just uses the pre-installed one from the base image.

SimBe195 (changes requested):

Atticus1806:
So the question to decide now is: Do we wait for t 2.14 or do we just stick with the potential issue in 2.13? I am fine with both, I would just not like to invest time in getting tf to compile.

curufinwe:
There is already an 2.14.0-rc1 version (2nd release candidate). Number of release candidates seems to vary between 2-4. My guess would be that we'll see a new release in 1-2 months. If you can wait that long I would wait. If you really need a new version and don't need conformers I would say go ahead and add&use this image.

Atticus1806:
I am fine with waiting.

Atticus1806:
So this now works. Compiled the newest RASR version. 
@curufinwe can you have a look again please? :)

SimBe195:
I can confirm that tf 2.14 does contain the fixes as seen here
[https://github.com/tensorflow/tensorflow/blob/v2.14.0/tensorflow/core/grappler/optimizers/remapper.cc#L1886]()
and here
[https://github.com/tensorflow/tensorflow/blob/v2.14.0/tensorflow/core/grappler/optimizers/remapper.cc#L4251]()

SimBe195 (approved):

SimBe195 (changes requested):
In this image there is a mismatch in that `/usr/bin/python3` links to `python3.11` but `/usr/bin/python3-config` links to `python3.10-config` so the binaries will be linked to the python 3.10 libraries during compilation which can cause issues. You have to either correct the `/usr/bin/python3-config` itself or change `python3-config` to `python3.11-config` in the `config/os-linux.make` file.

SimBe195 (approved):

curufinwe (approved):

PR #69, opened 2023-11-10
Add python bindings

DanEnergetics:
- Add LibRASR shared library
- Make a pip installable package
- RASR features available in Python module:
  - create Configuration object and read a RASR config file from disk
  - Pass configuration to AllophoneStateFsaBuilder to build weighted Automata
- Let Core::Configuration class have Core::Ref pointer to ResourceDataBase object


src/Python/Makefile
```diff
@@ -12,8 +12,11 @@ SUBDIRS 	=
 TARGETS		= libSprintPython.$(a) check$(exe)
 
 LIBPYTHON_O    = \
```
curufinwe, line 14: You need to add
```
ifdef MODULE_TENSORFLOW
CXXFLAGS += $(TF_CXXFLAGS)
LDFLAGS  += $(TF_LDFLAGS)
endif
```
to the Makefile (ideally after CCFLAGS)

curufinwe (changes requested):
We also need the following changes to the recently merged apptainer image:
```
diff --git a/apptainer/2023-11-08_tensorflow-2.14_v1/image.def b/apptainer/2023-11-08_tensorflow-2.14_v1/image.def
index 1ad19d7..6138f0c 100644
--- a/apptainer/2023-11-08_tensorflow-2.14_v1/image.def
+++ b/apptainer/2023-11-08_tensorflow-2.14_v1/image.def
@@ -10,6 +10,8 @@ Stage: build
         bison libxml2-dev libopenblas-dev libsndfile1-dev libcrypto++-dev libcppunit-dev \
         parallel xmlstarlet python3-lxml htop strace gdb sox python3-pip cmake ffmpeg vim
 
+    pip3 install pybind11==2.11.1
+
     # download the cache manager and place in /usr/local
     cd /usr/local
     git clone https://github.com/rwth-i6/cache-manager.git
diff --git a/apptainer/2023-11-08_tensorflow-2.14_v1/makefiles/config/os-linux.make b/apptainer/2023-11-08_tensorflow-2.14_v1/makefiles/config/os-linux.make
index cd7549d..a0a150d 100644
--- a/apptainer/2023-11-08_tensorflow-2.14_v1/makefiles/config/os-linux.make
+++ b/apptainer/2023-11-08_tensorflow-2.14_v1/makefiles/config/os-linux.make
@@ -144,6 +144,8 @@ LDFLAGS     += `${PYTHON_PATH}/bin/python3.11-config --ldflags --embed 2>/dev/nu
 LDFLAGS     += -Wl,-rpath -Wl,${PYTHON_PATH}/lib
 else
 INCLUDES    += `python3.11-config --includes 2>/dev/null`
+INCLUDES    += -I$(shell python3.11 -c 'import numpy as np; print(np.get_include())')
+INCLUDES    += `python3.11 -m pybind11 --includes 2>/dev/null`
 LDFLAGS     += `python3.11-config --ldflags --embed 2>/dev/null`
 # IF you want to use Python2 for whatever reason:
 # INCLUDES    += `pkg-config --cflags python`
```

src/Python/PybindModule.cc
```diff
@@ -0,0 +1,96 @@
...
+    INIT_MODULE(Lm);
+    INIT_MODULE(Signal);
+    INIT_MODULE(Speech);
+#ifdef MODULE_NN
+    INIT_MODULE(Nn);
+#endif
+#ifdef MODULE_ONNX
+    INIT_MODULE(Onnx);
+#endif
+#ifdef MODULE_STREAMING
```
curufinwe, line 60: This module is not part of github RASR -> remove

src/Python/PybindModule.cc
```diff
@@ -0,0 +1,96 @@
...
+#ifdef MODULE_TENSORFLOW
+#include <Tensorflow/Module.hh>
+#endif
+
+#include "AllophoneStateFsaBuilder.hh"
+#include "Configuration.hh"
+
+namespace py = pybind11;
+
+class _DummyApplication : Core::Application {
```
curufinwe, line 32: After some thought, I think it would be best to merge this class with it's duplicate in Tools/LibRASR and also move the PYBIND11_MODULE into Tools/LibRasr

curufinwe (changes requested):
Some last changes. Please also talk to @NurAd-Din and/or @hannah220 .

DanEnergetics:
I'm wondering if we want to be able to `make` with calling `pip install .` instead of or as an additional alternative to making it manually beforehand. This is possible with some setuptools magic.

JackTemaki:
> I'm wondering if we want to be able to `make` with calling `pip install .` instead of or as an additional alternative to making it manually beforehand. This is possible with some setuptools magic.

Something like here: https://github.com/flashlight/text/blob/main/setup.py would definitely be nice... The question is how much overhead this is to be stable and working in most cases.

DanEnergetics:
Yes that's what I was thinking about. I tried a similar implementation using the existing make system and it seems to work. The only thing is that we might want to pass arguments to the `pip install .` call in order to call `configure --i6 --disable-modules=...` or something like that.

curufinwe (approved):

JackTemaki:
```
/usr/bin/ld: ../../Python/libSprintPython.linux-x86_64-standard.a(AllophoneStateFsaBuilder.o): relocation R_X86_64_TPOFF32 against hidden symbol `_ZZN8pybind116handle15inc_ref_counterEmE7counter' can not be used when making a shared object
```

JackTemaki:
I can compile it with my image after some fixes (which are unrelated to this PR), run setuptools and access the package in IPython. I will do some testing together with @Marvin84 tomorrow.

JackTemaki (approved):
Can confirm that it works. There are some improvements possible but I guess this can all be future work.

JackTemaki:
@curufinwe as there is some extra code needed to convert from the FSA here to the FSA needed for fast baum welch? Is there already a plan to put this code somewhere? E.g. to i6_models? I am talking about something like this, which I found in Daniels setup:
```python
    def build_batch(self, seq_tags):
        def concat_fsas(a: Tuple, b: Tuple):
            edges = torch.from_numpy(np.int32(b[2])).reshape((3, b[1]))
            return (
                a[0] + [b[0]], # num states
                a[1] + [b[1]], # num edges
                torch.hstack([a[2], edges]), # edges
                torch.cat([a[3], torch.from_numpy(b[3])]), # weights
            )
        fsas = map(self.builder.build_by_segment_name, seq_tags)
        num_states, num_edges, all_edges, all_weights = reduce(
            concat_fsas, fsas, ([], [], torch.empty((3, 0), dtype=torch.int32), torch.empty((0,)))
        )
        num_edges = torch.tensor(num_edges, dtype=torch.int32)
        num_states = torch.tensor(num_states, dtype=torch.int32)

        cum_num_states = torch.cumsum(num_states, dim=0, dtype=torch.int32)
        state_offsets = torch.cat([torch.zeros((1,), dtype=torch.int32), cum_num_states[:-1]])
        start_end_states = torch.vstack([state_offsets, cum_num_states - 1])

        edge_seq_idxs = torch.repeat_interleave(num_edges)
        all_edges[:2,:] += torch.repeat_interleave(state_offsets, num_edges)
        all_edges = torch.vstack([all_edges, edge_seq_idxs])

        if self.tdp_scale != 1.0:
            all_weights *= self.tdp_scale

        return (
            cum_num_states[-1],
            all_edges.to(cuda_device),
            all_weights.to(cuda_device),
            start_end_states.to(cuda_device)
        )
    ```

DanEnergetics:
> is there already a plan to put this code somewhere? E.g. to i6_models? [...]
There is an updated version of this being pushed to the AppTek recipes. I discussed this shortly with @michelwi, who suggested to put it under i6_experiments. I'm also leaning towards i6_models

PR #71, opened 2024-03-15
Update build system from Make to CMake

SimBe195:
This PR adds a CMake compilation system as a replacement of the old Makefile system. The new CMake compilation has several advantages:

- CMakeLists are generally easier to read and understand (and thus to extend) compared to Makefiles.
- Locations of certain dependencies are detected automatically instead of having to be specified via an absolute path somewhere in one of the Makefiles (e.g. Python, Tensorflow, ONNX).
- The project is easier to parse by IDEs and it's easier to generate a `compile_commands.json` for C++ LSPs.
- Builds are out-of-source and thus easy to separate and clean up.
- Different build systems such as Ninja can be used to speed up compilation.
- In the Makefile system even without any code changes, recompilation builds all executables again. With the CMake-system it only builds what is necessary.

The mechanism of enabling/disabling modules works a bit different (and less hacky) now compared to the Makefile way: Previously, a `Modules.hh` file was generated containing a bunch of `#define MODULE_...` statements and included all over the code base to check whether certain modules are enabled or not. In the CMake system the modules are just added directly as compile definitions and this `Modules.hh` generation and including is not needed anymore.


curufinwe:
@SimBe195 There is still an issue with building the python package because librasr is now named differently:
I use the following commands to build RASR:
```
cmake -S . -B build
cmake --build build -j8
```
This results in the librasr.so being located in:
```
build/src/Tools/LibRASR/librasr.cpython-312-x86_64-linux-gnu.so
```
The name is probably automatically generated by the `pybind11_add_module` rule. The problem is that in setup.py we have hardcoded `src/Tools/LibRASR/librasr.so`.

Could you look into this? TY.
Bonus points if you add building of the python module to the github workflow.

curufinwe (approved):

larissakl (approved):

PR #73, opened 2024-07-22
Content for first pull request.

NurAd-Din:
This is the first pull request I'm making for the new python bindings. This is also the first pull request I've ever made so please give some guidance on whether I wish add more content per pull request. This request includes some smaller classes relating to the major lexicon class exposure.

### DIFF
--- a/src/Tools/LibRASR/PybindModule.cc
+++ b/src/Tools/LibRASR/PybindModule.cc
@@ -1,6 +1,35 @@
 #include <string>
 
 #include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include <pybind11/complex.h>
+#include <pybind11/functional.h>
+#include <pybind11/chrono.h>
+#include <pybind11/iostream.h>
+
+#include <Am/Module.hh>
+#include <Audio/Module.hh>
+#include <Bliss/CorpusDescription.hh>
+#include <Bliss/CorpusParser.hh>
+#include <Bliss/Lexicon.hh>
+#include <Core/Application.hh>
+#include <Flf/Module.hh>
+#include <Flow/Module.hh>
+#include <Lm/Module.hh>
+#include <Math/Module.hh>
+#include <Mm/Module.hh>
+#include <Modules.hh>
+#ifdef MODULE_NN
+#include <Nn/Module.hh>
+#endif
+#ifdef MODULE_ONNX
+#include <Onnx/Module.hh>
+#endif
+#include <Signal/Module.hh>
+#include <Speech/Module.hh>
+#ifdef MODULE_TENSORFLOW
+#include <Tensorflow/Module.hh>
+#endif
 
 #include <Python/AllophoneStateFsaBuilder.hh>
 #include <Python/Configuration.hh>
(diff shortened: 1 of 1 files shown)


JackTemaki:
> This is also the first pull request I've ever made so please give some guidance

Sure, here some hints:
 - change the title to reflect the actual changes / additions of your PR
 - Change the description text and add a few more details what you are actually adding, although in this case there might be no useful information
 - regarding the PR size: this is usually hard to determine. I would say here the addition is quite small, so if there is more useful things to add to that, I think you can make the PR a little larger. Just try to have this in blocks which cover related logic. As I am not familiar with RASR internals, it is hard for me to give guidance on what should be added at the same time.

src/Tools/LibRASR/PybindModule.cc
```diff
@@ -27,4 +56,52 @@ PYBIND11_MODULE(librasr, m) {
...
                      &AllophoneStateFsaBuilder::buildBySegmentName);
+
+    py::class_<Bliss::Symbol> symbol(m, "Symbol");
+    symbol
+    .def(py::init<>())
+    .def(py::init<const Bliss::Symbol&>())
+    .def("length", &Bliss::Symbol::length)
+    .def("__eq__", &Bliss::Symbol::operator==)
+    .def("__ne__", &Bliss::Symbol::operator!=)
+    .def("__bool__", [](const Bliss::Symbol &self) { return !self.operator!(); })
```
hannah220, line 67: For `__bool__` function, why not use operator `.def("__bool__", &Bliss::Symbol::operator bool)` ?
Then `to_bool` function is not needed.

src/Tools/LibRASR/PybindModule.cc
```diff
@@ -27,4 +56,52 @@ PYBIND11_MODULE(librasr, m) {
...
+    .def("to_bool", &Bliss::Symbol::operator bool)
+    .def("to_string", &Bliss::Symbol::operator Bliss::Symbol::String)
+    .def("str", &Bliss::Symbol::str)
```
hannah220, line 70: How about changing the function name `str` -> `to_cstring` to make it comparable to `to_string`

src/Tools/LibRASR/PybindModule.cc
```diff
@@ -27,4 +56,52 @@ PYBIND11_MODULE(librasr, m) {
...
+
+    py::class_<Bliss::Symbol::Hash>(symbol, "Hash")
+    .def("__call__", &Bliss::Symbol::Hash::operator());
+
+    py::class_<Bliss::Symbol::Equality>(symbol, "Equality")
+    .def("__call__", &Bliss::Symbol::Equality::operator());
+
+    py::class_<Bliss::OrthographicFormList>(m, "OrthographicFormList")
+    .def(py::init<const Bliss::Symbol*, const Bliss::Symbol*>())
+    .def(py::init<>())
```
hannah220, line 81: Below, `Bliss::SymbolSequence<Bliss::OrthographicForm>` is `Bliss::OrthographicFormList`

src/Tools/LibRASR/PybindModule.cc
```diff
@@ -27,4 +56,52 @@ PYBIND11_MODULE(librasr, m) {
...
+    .def(py::init<const Bliss::SymbolSequence<Bliss::OrthographicForm>&>())
+    .def("valid", &Bliss::SymbolSequence<Bliss::OrthographicForm>::valid)
+    .def("size", &Bliss::SymbolSequence<Bliss::OrthographicForm>::size)
+    .def("length", &Bliss::SymbolSequence<Bliss::OrthographicForm>::length)
+    .def("is_epsilon", &Bliss::SymbolSequence<Bliss::OrthographicForm>::isEpsilon)
+    .def("front", &Bliss::SymbolSequence<Bliss::OrthographicForm>::front, py::return_value_policy::reference_internal)
+    .def("begin", &Bliss::SymbolSequence<Bliss::OrthographicForm>::begin, py::return_value_policy::reference_internal)
+    .def("end", &Bliss::SymbolSequence<Bliss::OrthographicForm>::end, py::return_value_policy::reference_internal)
+    .def("__getitem__", &Bliss::OrthographicFormList::operator[]);
```
hannah220, line 90: Should `__getitem__` also have `py::return_value_policy::reference_internal` ?

src/Tools/LibRASR/PybindModule.cc
```diff
@@ -27,4 +56,52 @@ PYBIND11_MODULE(librasr, m) {
...
+
+    py::class_<Bliss::SyntacticTokenSequence>(m, "SyntacticTokenSequence")
+    .def(py::init<>())
```
hannah220, line 93: Below, `SymbolSequence<const SyntacticToken*>` is `SyntacticTokenSequence`

src/Tools/LibRASR/PybindModule.cc
```diff
@@ -27,4 +56,52 @@ PYBIND11_MODULE(librasr, m) {
...
+    .def(py::init<const Bliss::SymbolSequence<const Bliss::SyntacticToken*>&>())
+    .def("valid", &Bliss::SymbolSequence<const Bliss::SyntacticToken*>::valid)
+    .def("size", &Bliss::SymbolSequence<const Bliss::SyntacticToken*>::size)
+    .def("length", &Bliss::SymbolSequence<const Bliss::SyntacticToken*>::length)
+    .def("is_epsilon", &Bliss::SymbolSequence<const Bliss::SyntacticToken*>::isEpsilon)
+    .def("front", &Bliss::SymbolSequence<const Bliss::SyntacticToken*>::front, py::return_value_policy::reference_internal)
```
hannah220, line 99: `begin`, `end` functions are missing

hannah220 (changes requested):

NurAd-Din:
I've changed the request to include all Lexicon class exposure for approval. I've also implemented some of the requested changes.

src/Tools/LibRASR/PybindModule.cc
hannah220: By Nour: no need for reference_internal because it returns copy

hannah220:
Some class has a single default constructor `.def(py::init<>())`, some don't have a constructor at all. 
If the C++ class has an implicit or explicitly defined default constructor (e.g., no arguments and does nothing special), and no other constructors, pybind11 will automatically expose it, and you don’t need to define `.def(py::init<>())`.
Can you make it consistent by only defining `.def(py::init<>())` when it's necessary?

src/Tools/LibRASR/PybindModule.cc
```diff
@@ -27,4 +116,286 @@ PYBIND11_MODULE(librasr, m) {
...
+
+    py::class_<Bliss::Symbol::Hash>(symbol, "Hash")
+    .def("__call__", &Bliss::Symbol::Hash::operator());
+
+    py::class_<Bliss::Symbol::Equality>(symbol, "Equality")
+    .def("__call__", &Bliss::Symbol::Equality::operator());
+
+    py::class_<Bliss::OrthographicFormList>(m, "OrthographicFormList")
+    .def(py::init<const Bliss::Symbol*, const Bliss::Symbol*>())
+    .def(py::init<>())
```
hannah220, line 140: Could you define this `.def(py::init<>())` at the beginning? Just so it looks nicer

src/Tools/LibRASR/PybindModule.cc
```diff
@@ -27,4 +116,286 @@ PYBIND11_MODULE(librasr, m) {
...
+    .def("has_pronunciation", &Bliss::Lemma::hasPronunciation)
+    .def("has_syntactic_token_sequence", &Bliss::Lemma::hasSyntacticTokenSequence)
+    .def("syntactic_token_sequence", &Bliss::Lemma::syntacticTokenSequence, py::return_value_policy::reference_internal)
+    .def("has_evaluation_token_sequence", &Bliss::Lemma::hasEvaluationTokenSequence)
+    .def("num_evaluation_token_sequences", &Bliss::Lemma::nEvaluationTokenSequences);
+
+    py::class_<Bliss::Lemma::PronunciationIterator>(lemma, "PronunciationIterator")
+    .def(py::init<>())
+    .def("to_lemma_pronunciation", &Bliss::Lemma::PronunciationIterator::operator const Bliss::LemmaPronunciation*)
+    .def("increment", &Bliss::Lemma::PronunciationIterator::operator++, py::return_value_policy::reference_internal);
```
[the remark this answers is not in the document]
hannah220: That's a good idea, let's also change so in other places

hannah220 (changes requested):

NurAd-Din:
Previously requested changes implemented.

src/Tools/LibRASR/PybindModule.cc
```diff
@@ -27,4 +103,285 @@ PYBIND11_MODULE(librasr, m) {
...
+    .def("class_emission_score", &Bliss::SyntacticToken::classEmissionScore);
+
+    py::class_<Bliss::Letter, Bliss::Token>(m, "Letter");
+
+    py::class_<Bliss::Phoneme, Bliss::Token>(m, "Phoneme")
+    .def("set_context_dependent", &Bliss::Phoneme::setContextDependent)
+    .def("is_context_dependent", &Bliss::Phoneme::isContextDependent)
+    .def_readonly_static("term", &Bliss::Phoneme::term);
+
+    py::class_<Bliss::PhonemeInventory, Core::Ref<Bliss::PhonemeInventory>>(m, "PhonemeInventory")
```
hannah220, line 215: When the parent class is `Core::ReferenceCounted`, you define Core::Ref<NameOfChildClass>, in this case `Core::Ref<Bliss::PhonemeInventory>`. 
Why is that?

NurAd-Din: In these cases Core::References allows for the use of Core::Ref<> as a reference counter. It serves no purpose as a parent class other than that. So we directly specify to pybind11 that these objects are to be created and handled in Core::Ref. Core::Ref itself is specified to pybind11 as a sort of smart pointer before I started exposing classes.

hannah220 (approved):

curufinwe (approved):

PR #74, opened 2024-09-20
Backport AppTek changes to LM + LM lookahead code

curufinwe:
Originally I wanted to only backport some fixes I made to the cheating LM, s.t. you can also do search-error experiments, but this escalated to many more changes in the LM codebase + some search related stuff. I tested that everything still compiles (in an appropriate singularity image), but it would be good if one / multiple of you could check if they get the same recognition results as before.

### DIFF
--- a/src/Bliss/SyntacticTokenMap.hh
+++ b/src/Bliss/SyntacticTokenMap.hh
@@ -62,6 +62,16 @@ public:
         verify_(s->id() < store_.size());
         return store_[s->id()];
     }
+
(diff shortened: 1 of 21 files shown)


Marvin84:
I tested both 4gram and transformer LM decoding and got the same WER. Could we merge this PR? @curufinwe @SimBe195

Marvin84 (approved):

SimBe195 (approved):

PR #77, opened 2024-10-29
Basic text lexicon parser

larissakl:
Support for simple text lexica instead of XML-files.
This is meant for unconstrained search for which the XML-format is unnecessarily complicated.
Instead of an XML-lexicon, one can use a txt-file containing one label per line.
For example
```
A
B
C
```
will be parsed to a Bliss Lexicon equivalent to 
```
<?xml version="1.0" ?>
<lexicon>
  <phoneme-inventory>
    <phoneme>
      <symbol>A</symbol>
      <variation>none</variation>
    </phoneme>
    <phoneme>
      <symbol>B</symbol>
      <variation>none</variation>
    </phoneme>
    <phoneme>
      <symbol>C</symbol>
      <variation>none</variation>
    </phoneme>
    <phoneme>
 </phoneme-inventory>
  <lemma>
    <orth>A</orth>
    <phon>A</phon>
  </lemma>
  <lemma>
    <orth>B</orth>
    <phon>B</phon>
  </lemma>
  <lemma>
    <orth>C</orth>
    <phon>C</phon>
  </lemma>
</lexicon>
```

### DIFF
--- a/src/Bliss/Lexicon.cc
+++ b/src/Bliss/Lexicon.cc
@@ -142,7 +142,7 @@ const Core::Choice Lexicon::lexiconTypeChoice(
         "xml", xmlLexicon,
         Core::Choice::endMark());
 
-const Core::ParameterChoice Lexicon::lexiconTypeParam(
+const Core::ParameterChoice Lexicon::paramLexiconType(
         "type", &Lexicon::lexiconTypeChoice, "type of the lexicon file", xmlLexicon);
 
 void Lexicon::load(const std::string& filename) {
@@ -152,20 +152,20 @@ void Lexicon::load(const std::string& filename) {
     else
         warning("could not derive md5 sum from file '%s'", filename.c_str());
     std::unique_ptr<LexiconParser> parser;
-    Core::Choice::Value lexType = Lexicon::lexiconTypeParam(config);
-    switch (Lexicon::LexiconType(lexType)) {
+    switch (paramLexiconType(config)) {
     // text-based lexicon
-    case Lexicon::LexiconType::txtLexicon:
+    case LexiconType::txtLexicon:
         parser = std::make_unique<TextLexiconParser>(this);
         break;
     // xml-based lexicon
-    case Lexicon::LexiconType::xmlLexicon:
+    case LexiconType::xmlLexicon:
         parser = std::make_unique<XmlLexiconParser>(config, this);
         break;
     }
     log("reading lexicon from file") << " \"" << filename << "\" ...";
-    if (!parser->parseFile(filename))
+    if (!parser->parseFile(filename)) {
         error("Error while reading lexicon file.");
+    }
     log("dependency value: ") << dependency_.value();
 }
 
--- a/src/Bliss/Lexicon.cc
+++ b/src/Bliss/Lexicon.cc
@@ -137,8 +137,8 @@ Lexicon::~Lexicon() {
 }
 
 const Core::Choice Lexicon::lexiconTypeChoice(
-        "txt", txtLexicon,
-        "text", txtLexicon,
+        "vocab-txt", vocabTxtLexicon,
+        "vocab-text", vocabTxtLexicon,
         "xml", xmlLexicon,
         Core::Choice::endMark());
 
@@ -147,14 +147,15 @@ const Core::ParameterChoice Lexicon::paramLexiconType(
 
 void Lexicon::load(const std::string& filename) {
     Core::MD5 md5;
-    if (md5.updateFromFile(filename))
+    if (md5.updateFromFile(filename)) {
         dependency_.setValue(md5);
-    else
+    } else {
         warning("could not derive md5 sum from file '%s'", filename.c_str());
+    }
     std::unique_ptr<LexiconParser> parser;
     switch (paramLexiconType(config)) {
     // text-based lexicon
-    case LexiconType::txtLexicon:
+    case LexiconType::vocabTxtLexicon:
         parser = std::make_unique<TextLexiconParser>(this);
         break;
     // xml-based lexicon
--- a/src/Bliss/Lexicon.cc
+++ b/src/Bliss/Lexicon.cc
@@ -142,9 +142,15 @@ void Lexicon::load(const std::string& filename) {
         dependency_.setValue(md5);
     else
         warning("could not derive md5 sum from file '%s'", filename.c_str());
-    LexiconParser parser(config, this);
+    std::unique_ptr<LexiconParser> parser;
+    // text-based lexicon
+    if (filename.compare(filename.length()-4, 4, ".txt") == 0)
+        parser = std::make_unique<TextLexiconParser>(this);
+    // xml-based lexicon
+    else
+        parser = std::make_unique<XmlLexiconParser>(config, this);
     log("reading lexicon from file") << " \"" << filename << "\" ...";
-    if (parser.parseFile(filename.c_str()) != 0)
+    if (parser->parseFile(filename) != 0)
         error("Error while reading lexicon file.");
     log("dependency value: ") << dependency_.value();
 }
--- a/src/Bliss/Lexicon.cc
+++ b/src/Bliss/Lexicon.cc
@@ -156,7 +156,7 @@ void Lexicon::load(const std::string& filename) {
     switch (paramLexiconType(config)) {
     // text-based lexicon
     case LexiconType::vocabTxtLexicon:
-        parser = std::make_unique<TextLexiconParser>(this);
+        parser = std::make_unique<VocabTextLexiconParser>(this);
         break;
     // xml-based lexicon
     case LexiconType::xmlLexicon:
--- a/src/Bliss/Lexicon.cc
+++ b/src/Bliss/Lexicon.cc
@@ -136,6 +136,15 @@ Lexicon::~Lexicon() {
     delete internal_;
 }
 
+const Core::Choice Lexicon::lexiconTypeChoice(
+        "txt", txtLexicon,
+        "text", txtLexicon,
+        "xml", xmlLexicon,
+        Core::Choice::endMark());
+
+const Core::ParameterChoice Lexicon::lexiconTypeParam(
+        "type", &Lexicon::lexiconTypeChoice, "type of the lexicon file", xmlLexicon);
+
 void Lexicon::load(const std::string& filename) {
     Core::MD5 md5;
     if (md5.updateFromFile(filename))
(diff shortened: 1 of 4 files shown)


src/Bliss/LexiconParser.hh
```diff
@@ -96,17 +99,26 @@ public:
...
  * Parser for Bliss lexicon files.
  * This class implements parsing of the lexicon XML format
  * described in <a href="../../doc/Lexicon.pdf">Lexicon File
  * Format Reference</a>.  It is normally not used directly but
  * through Lexicon.
  */
-
-class LexiconParser : public Core::XmlSchemaParser {
+class XmlLexiconParser : public virtual LexiconParser, public Core::XmlSchemaParser {
     typedef Core::XmlSchemaParser Precursor;
```
SimBe195, line 120: Avoid the `Precursor` definition with multiple inheritance like this since it's not clear later which parent class `Precursor` refers to later without checking back.

src/Bliss/LexiconParser.cc
```diff
@@ -379,12 +379,100 @@ void LexiconParser::loadWhitelist(const Core::Configuration& config, Core::Strin
...
+void TextLexiconParser::createPhoneme(const std::string line) {
+    std::string symbol(line);
+    stripWhitespace(symbol);        // in case there are any unintentional whitespaces
+    suppressTrailingBlank(symbol);
+
+    // check if phoneme was already added (if one label appears more than once)
+    const Phoneme* test = phonemeInventory_->phoneme(symbol);
+    if (test != 0) {
+        return;
+    }
```
SimBe195, line 448:
```suggestion
    if (phonemeInventory_->phoneme(symbol)) {
        return;
    }
```

src/Bliss/LexiconParser.cc
```diff
@@ -379,12 +379,100 @@ void LexiconParser::loadWhitelist(const Core::Configuration& config, Core::Strin
...
+
+    // create a new phoneme
+    Phoneme* currentPhoneme_ = phonemeInventory_->newPhoneme();
+    // set symbol
+    phonemeInventory_->assignSymbol(currentPhoneme_, symbol);
+    // set variation to none
+    currentPhoneme_->setContextDependent(false);
```
SimBe195, line 455:
```suggestion
    Phoneme* newPhoneme_ = phonemeInventory_->newPhoneme();
    // set symbol
    phonemeInventory_->assignSymbol(newPhoneme_, symbol);
    // set variation to none
    newPhoneme_->setContextDependent(false);
```

src/Bliss/LexiconParser.cc
```diff
@@ -379,12 +379,100 @@ void LexiconParser::loadWhitelist(const Core::Configuration& config, Core::Strin
...
+void TextLexiconParser::createLemma(const std::string line) {
+    std::string symbol(line);
+    stripWhitespace(symbol);        // in case there are any unintentional whitespaces
+    suppressTrailingBlank(symbol);
+
+    // check if lemma was already added (if one label appears more than once)
+    const Lemma* test = lexicon_->lemma(symbol);
+    if (test != 0) {
+        return;
+    }
```
SimBe195, line 468:
```suggestion
    if (lexicon_->lemma(symbol)) {
        return;
    }
```

src/Bliss/LexiconParser.cc
```diff
@@ -379,12 +379,100 @@ void LexiconParser::loadWhitelist(const Core::Configuration& config, Core::Strin
...
+
+    // create a new lemma
+    Lemma* currentLemma_ = lexicon_->newLemma();
+    // set orth
+    lexicon_->setOrthographicForms(currentLemma_, {symbol});
+    // set phon
+    Pronunciation* pron = lexicon_->getPronunciation(symbol);
+    lexicon_->addPronunciation(currentLemma_, pron);
+    lexicon_->setDefaultLemmaName(currentLemma_);
```
SimBe195, line 477:
```suggestion
    Lemma* newLemma_ = lexicon_->newLemma();
    // set orth
    lexicon_->setOrthographicForms(newLemma_, {symbol});
    // set phon
    Pronunciation* pron = lexicon_->getPronunciation(symbol);
    lexicon_->addPronunciation(newLemma_, pron);
    lexicon_->setDefaultLemmaName(newLemma_);
```

src/Bliss/LexiconParser.hh
```diff
@@ -96,17 +99,26 @@ public:
     virtual void characters(const char*, int){};
 };
 
+/*
+ * Base lexicon parser class
+ */
+class LexiconParser {
+public:
+    virtual ~LexiconParser() {}
+    virtual int parseFile(const std::string& filename) = 0;
```
SimBe195, line 108: I'm not sure why `parseFile` is supposed to return an `int` even though the return value is only ever used as a boolean, right? Since you are touching the code anyway, I would convert the return type to `bool`.

SimBe195 (changes requested):

src/Bliss/LexiconParser.cc
```diff
@@ -379,12 +379,100 @@ void LexiconParser::loadWhitelist(const Core::Configuration& config, Core::Strin
...
+    while (std::getline(file_first_pass, line)) {
+        if (line.empty()) continue;
+        createPhoneme(line);
+    }
+    file_first_pass.close();
+
+    // set the phoneme inventory
+    lexicon_->setPhonemeInventory(Core::ref(phonemeInventory_));
+
+    // second pass: create the lemmata in the lexicon
```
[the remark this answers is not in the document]
larissakl: I think this is not directly possible. The phoneme inventory needs to be set before pronunciation for a lemma can be set because `Lexicon::getPronunciation()` requires the phoneme inventory.
 Instead, we could pass the file once, set the phonemes and collect the labels as string in a vector and then do another pass over this vector to set the lemmata. This also requires two passes, but only one pass over the file. What do you think?

src/Bliss/LexiconParser.hh
larissakl: This was left from the original code and I tried to avoid touching that... So in the constructor, I can just use `XmlSchemaParser(c)` instead of `Precursor(c)`, right?

src/Bliss/Lexicon.cc
```diff
@@ -142,9 +142,15 @@ void Lexicon::load(const std::string& filename) {
         dependency_.setValue(md5);
     else
         warning("could not derive md5 sum from file '%s'", filename.c_str());
-    LexiconParser parser(config, this);
+    std::unique_ptr<LexiconParser> parser;
+    // text-based lexicon
+    if (filename.compare(filename.length()-4, 4, ".txt") == 0)
```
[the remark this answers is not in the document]
SimBe195: Right, that's what I meant. You can search for instances of `ParameterChoice` in other places in the RASR code to see how this is done.

src/Bliss/LexiconParser.cc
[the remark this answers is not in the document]
SimBe195: You also don't need a vector I believe. The `PhonemeInventory` class provides a function `phonemes()` which returns a pair of iterators that you can use for the second pass.

src/Bliss/LexiconParser.hh
SimBe195: Originally `XmlSchemaParser` was also the only parent but now `Precursor` is ambiguous.

src/Bliss/LexiconParser.cc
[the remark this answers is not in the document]
larissakl: Ah yes, good idea. Then I will change the `createLemma()` method such that it iterates over the phonemes and creates the lemmata and call it only once in `parseFile()`.

src/Bliss/LexiconParser.hh
larissakl: Please let me know if this is correct now

src/Bliss/Lexicon.cc
```diff
@@ -136,15 +136,35 @@ Lexicon::~Lexicon() {
     delete internal_;
 }
 
+const Core::Choice Lexicon::lexiconTypeChoice(
+        "txt", txtLexicon,
+        "text", txtLexicon,
```
SimBe195, line 141: Maybe call it a different name like `vocab-txt` or something so that the type suggests that this is just a file with a list of labels. If it's called `txt` and later someone wants to build a full text-based pronunciation lexicon parser the naming will become difficult/confusing. Such a name change could also be applied to the class names.

src/Bliss/Lexicon.cc
```diff
@@ -136,15 +136,35 @@ Lexicon::~Lexicon() {
...
+        "xml", xmlLexicon,
+        Core::Choice::endMark());
+
+const Core::ParameterChoice Lexicon::lexiconTypeParam(
+        "type", &Lexicon::lexiconTypeChoice, "type of the lexicon file", xmlLexicon);
```
SimBe195, line 146:
```suggestion
const Core::ParameterChoice Lexicon::paramLexiconType(
        "type", &Lexicon::lexiconTypeChoice, "type of the lexicon file", xmlLexicon);
```

src/Bliss/Lexicon.cc
```diff
@@ -136,15 +136,35 @@ Lexicon::~Lexicon() {
...
 void Lexicon::load(const std::string& filename) {
     Core::MD5 md5;
     if (md5.updateFromFile(filename))
         dependency_.setValue(md5);
     else
         warning("could not derive md5 sum from file '%s'", filename.c_str());
-    LexiconParser parser(config, this);
+    std::unique_ptr<LexiconParser> parser;
+    Core::Choice::Value lexType = Lexicon::lexiconTypeParam(config);
+    switch (Lexicon::LexiconType(lexType)) {
```
SimBe195, line 156:
```suggestion
    switch (paramLexiconType(config)) {
```

src/Bliss/Lexicon.cc
```diff
@@ -136,15 +136,35 @@ Lexicon::~Lexicon() {
...
+    // text-based lexicon
+    case Lexicon::LexiconType::txtLexicon:
```
SimBe195, line 158:
```suggestion
    case LexiconType::txtLexicon:
```

src/Bliss/Lexicon.cc
```diff
@@ -136,15 +136,35 @@ Lexicon::~Lexicon() {
...
+        parser = std::make_unique<TextLexiconParser>(this);
+        break;
+    // xml-based lexicon
+    case Lexicon::LexiconType::xmlLexicon:
```
SimBe195, line 162:
```suggestion
    case LexiconType::xmlLexicon:
```

src/Bliss/Lexicon.cc
```diff
@@ -136,15 +136,35 @@ Lexicon::~Lexicon() {
...
+        parser = std::make_unique<XmlLexiconParser>(config, this);
+        break;
+    }
     log("reading lexicon from file") << " \"" << filename << "\" ...";
-    if (parser.parseFile(filename.c_str()) != 0)
+    if (!parser->parseFile(filename))
         error("Error while reading lexicon file.");
```
SimBe195, line 168:
```suggestion
    if (!parser->parseFile(filename)) {
        error("Error while reading lexicon file.");
    }
```

src/Bliss/LexiconParser.cc
```diff
@@ -379,12 +379,90 @@ void LexiconParser::loadWhitelist(const Core::Configuration& config, Core::Strin
...
+
+    // set the phoneme inventory
+    lexicon_->setPhonemeInventory(Core::ref(phonemeInventory_));
+    // iterate over the phonemes in the inventory to create the lemmata in the lexicon
+    createLemmata();
+    return true;
+}
+
+// helper function to handle one label and create a corresponding phoneme
+void TextLexiconParser::createPhoneme(const std::string line) {
```
SimBe195, line 428:
```suggestion
void TextLexiconParser::createPhoneme(const std::string& line) {
```

SimBe195 (changes requested):

PR #78, opened 2024-10-31
Add Core::CollapsedVector

SimBe195:
Vector that is automatically collapsed down to one element if all entries have the same value
From the outside it behaves like a normal vector. Whether it is collapsed or not only matters internally.

Example:

```
CollapsedVector<int> v;  // Internal data of v is [] with logical size 0
v.push_back(5);  // Normal push. Internal data of v is [5] with logical size 1
v.push_back(5);  // v stays collapsed. Internal data of v is [5] with logical size 2
std::cout << v.size() << std::endl;  // Prints "2"
std::cout << v[1] << std::endl;  // Prints "5"
v.push_back(6);  // v is expanded because the new value is different. Internal data of v is [5, 5, 6] with logical size 3
std::cout << v[2] << std::endl;  // Prints "6"
v.clear();  // Internal data of v is [] with logical size 0.
```

### DIFF
--- a/src/Core/CollapsedVector.hh
+++ b/src/Core/CollapsedVector.hh
@@ -1,0 +1,130 @@
+/** Copyright 2024 RWTH Aachen University. All rights reserved.
+ *
+ *  Licensed under the RWTH ASR License (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.hltpr.rwth-aachen.de/rwth-asr/rwth-asr-license.html
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#ifndef COLLAPSED_VECTOR_HH
+#define COLLAPSED_VECTOR_HH
+
+#include <stdexcept>
+#include <vector>
+
+namespace Core {
+
+/*
+ * Vector that is automatically collapsed down to one element if all entries have the same value
+ * From the outside it behaves like a normal vector. Whether it is collapsed or not only matters internally.
+ *
+ * Example:
+ *
+ * CollapsedVector<int> v;  // Internal data of v is [] with logical size 0
+ * v.push_back(5);  // Normal push. Internal data of v is [5] with logical size 1
+ * v.push_back(5);  // v stays collapsed. Internal data of v is [5] with logical size 2
+ * std::cout << v.size() << std::endl;  // Prints "2"
+ * std::cout << v[1] << std::endl;  // Prints "5"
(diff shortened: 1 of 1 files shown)


src/Core/CollapsedVector.hh
```diff
@@ -0,0 +1,130 @@
...
+ * std::cout << v[2] << std::endl;  // Prints "6"
+ * v.clear();  // Internal data of v is [] with logical size 0.
+ */
+template<typename T>
+class CollapsedVector {
+public:
+    CollapsedVector();
+    CollapsedVector(size_t size, const T& value);
+
+    bool     isCollapsed() const;
```
larissakl: Is this still needed? It's not implemented below.

larissakl: Oh sorry, you were faster :)

larissakl (approved):

Marvin84 (approved):

src/Core/CollapsedVector.hh
```diff
@@ -0,0 +1,129 @@
...
+CollapsedVector<T>::CollapsedVector()
+        : data_(), logicalSize_(0ul) {}
+
+template<typename T>
+CollapsedVector<T>::CollapsedVector(size_t size, const T& value)
+        : data_(1ul, value), logicalSize_(size) {}
+
+template<typename T>
+void CollapsedVector<T>::push_back(const T& value) {
+    if (data_.empty() or data_.size() > 1ul) {
```
curufinwe, line 72: This can be shortened to `data_.size != 1`

src/Core/CollapsedVector.hh
```diff
@@ -0,0 +1,129 @@
...
+
+    ++logicalSize_;  // Logical size always increases
+}
+
+template<typename T>
+const T& CollapsedVector<T>::operator[](size_t idx) const {
+    if (data_.size() == 1ul) {  // Vector may be collapsed
+        return front();
+    }
+    return data_.operator[](idx);
```
curufinwe, line 95: no need to explicitly call `operator[]`, just write `data_[idx]`

src/Core/CollapsedVector.hh
```diff
@@ -0,0 +1,129 @@
...
+}
+
+template<typename T>
+const T& CollapsedVector<T>::at(size_t idx) const {
+    if (idx >= logicalSize_) {
+        throw std::out_of_range("Trying to access illegal index of CollapsedVector");
+    }
+    return operator[](idx);
```
curufinwe, line 103: Alternative: `(*this)[idx]`. But feel free to also keep your version if you like it better.

curufinwe (changes requested):
In addition to inline comments, please also add `inline` to the function definitions, s.t. the linker does not treat it as an error if two object files with the same symbol are linked together.

src/Core/CollapsedVector.hh
SimBe195: Right, I think this is probably because initially the class inherited from `std::vector` instead of having it as a member so later I replaced all occurrences of `this->` with `data_.`

curufinwe (approved):

PR #79, opened 2024-10-31
Add Core::FIFOCache

SimBe195:
Cache based on an unordered_map with a maximum size such that when the cache is full and
a new item is trying to be added, the oldest item in the cache gets removed.
Note: oldest is determined by order of insertion, not order of last access.

Example:
```
FIFOCache<int, std::string> cache(2);  // Cache has room for two items
cache.put(1, "one");  // Internal data: {1: "one"}
cache.put(2, "two");  // Internal data: {1: "one", 2: "two"}
cache.put(3, "three");  // Oldest element is deleted since max size was reached. Internal data: {2: "two", 3: "three"}
```


larissakl (approved):

src/Core/FIFOCache.hh
```diff
@@ -0,0 +1,131 @@
...
+// Insert or update a key-value pair (oldest inserted elements get removed first)
+template<typename Key, typename Value, typename Hash, typename KeyEqual>
+void FIFOCache<Key, Value, Hash, KeyEqual>::put(const Key& key, const Value& value) {
+    auto it = cacheMap_.find(key);
+    if (it != cacheMap_.end()) {
+        // Key exists; update the value but don't change the order in the list
+        it->second = value;
+    }
+    else {  // Key doesn't exist yet
+
```
curufinwe, line 83: superfluous newline

src/Core/FIFOCache.hh
```diff
@@ -0,0 +1,131 @@
...
+            // Next oldest element becomes new oldest
+            oldestElementPos_ = (oldestElementPos_ + 1ul) % maxSize_;  // Wrap position around at the end of the vector
+        }
+    }
+}
+
+template<typename Key, typename Value, typename Hash, typename KeyEqual>
+Value& FIFOCache<Key, Value, Hash, KeyEqual>::get(const Key& key) {
+    return cacheMap_.at(key);
```
curufinwe, line 104: This throws an exception if the key is not present. As nobody ever catches these it will probably terminate the application. I think it would be better to return `std::optional<Value&>` and for an interface the default-constructs the value add an `operator[]` function.

For the version returning `std::optional<Value&>` we could also have a const version of the function.

curufinwe (changes requested):

curufinwe:
PS: same comment about adding `inline` as in https://github.com/rwth-i6/rasr/pull/78

src/Core/FIFOCache.hh
SimBe195: C++ unfortunately doesn't support optional references so `std::optional<Value&>` doesn't work . I have implemented this now by using `std::optional<std::reference_wrapper<Value>>`. It could also be done by using `Value*` as return value instead and returning `nullptr` when the key is not found.

src/Core/FIFOCache.hh
```diff
@@ -0,0 +1,158 @@
...
+            oldestElementPos_ = (oldestElementPos_ + 1ul) % maxSize_;  // Wrap position around at the end of the vector
+        }
+    }
+}
+
+template<typename Key, typename Value, typename Hash, typename KeyEqual>
+inline std::optional<std::reference_wrapper<Value>> FIFOCache<Key, Value, Hash, KeyEqual>::get(const Key& key) {
+    auto it = cacheMap_.find(key);
+    if (it != cacheMap_.end()) {
+        return std::reference_wrapper<Value>(it->second);
```
curufinwe, line 109: Here the "recommended" way would be to use `std::ref()`: https://en.cppreference.com/w/cpp/utility/functional/ref
Also in the function below.

PR #80, opened 2024-10-31
Add LabelScorer base class

SimBe195:
Abstract base class for scoring tokens within an ASR search algorithm.

This class provides an interface for different types of label scorers in an ASR system. Label Scorers compute the scores of tokens based on input features and a scoring context. Children of this base class should represent various ASR model architectures and cover a wide range of possibilities such as CTC, transducer, AED or other models.

The usage is intended as follows:

- Before or during the search, features can be added
- At the beginning of search, `getInitialScoringContext` should be called and used for the first hypotheses
- For a given hypothesis in search, its search context together with a successor token and transition type are packed into a request and scored via `getScoreWithTime`. This also returns the timestamp of the successor.
  - Note: The scoring function may return no value, in this case it is not ready yet and needs more input features.
  - Note: There is also the function `getScoresWithTimes` which can handle an entire batch of requests at once and might be implemented more efficiently (e.g. using batched model forwarding).
- For all hypotheses that survive pruning, the LabelScorer can compute a new scoring context that extends the previous scoring context of that hypothesis with a given successor token. This new scoring context can then be used as context in subsequent search steps.
- After all features have been passed, the `signalNoMoreFeatures` function is called to inform the label scorer that it doesn't need to wait for more features and can score as much as possible. This is especially important when the label scorer internally uses an encoder or window with right context.
- When all necessary scores for the current segment have been computed, the `reset` function is called to clean up any internal data (e.g. feature buffer) or reset flags of the LabelScorer. Afterwards it is ready to receive features for the next segment.

Each concrete subclass internally implements a concrete type of scoring context which the outside search algorithm is agnostic to. Depending on the model, this scoring context can consist of things like the current timestep, a label history, a hidden state or other values.

### DIFF
--- a/src/Nn/LabelScorer/LabelScorer.hh
+++ b/src/Nn/LabelScorer/LabelScorer.hh
@@ -1,0 +1,144 @@
+/** Copyright 2024 RWTH Aachen University. All rights reserved.
+ *
+ *  Licensed under the RWTH ASR License (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.hltpr.rwth-aachen.de/rwth-asr/rwth-asr-license.html
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#ifndef LABEL_SCORER_HH
+#define LABEL_SCORER_HH
+
+#include <Core/CollapsedVector.hh>
+#include <Core/Component.hh>
+#include <Core/Configuration.hh>
+#include <Core/Parameter.hh>
+#include <Core/ReferenceCounting.hh>
+#include <Core/Types.hh>
+#include <Flow/Timestamp.hh>
+#include <Flow/Vector.hh>
+#include <Mm/FeatureScorer.hh>
+#include <Nn/Types.hh>
+#include <Search/Types.hh>
+#include <Speech/Feature.hh>
+#include <Speech/Types.hh>
+#include <optional>
+#include "ScoringContext.hh"
+
+namespace Nn {
+
+typedef Search::Score                Score;
+typedef Flow::Vector<f32>            FeatureVector;
+typedef Flow::DataPtr<FeatureVector> FeatureVectorRef;
+
+/*
+ * Abstract base class for scoring tokens within an ASR search algorithm.
+ *
+ * This class provides an interface for different types of label scorers in an ASR system.
+ * Label Scorers compute the scores of tokens based on input features and a scoring context.
+ * Children of this base class should represent various ASR model architectures and cover a
+ * wide range of possibilities such as CTC, transducer, AED or other models.
+ *
+ * The usage is intended as follows:
+ *  - Before or during the search, features can be added
+ *  - At the beginning of search, `getInitialScoringContext` should be called
+ *    and used for the first hypotheses
+ *  - For a given hypothesis in search, its search context together with a successor token and
+ *    transition type are packed into a request and scored via `getScoreWithTime`. This also returns
+ *    the timestamp of the successor.
+ *    Note: The scoring function may return no value, in this case it is not ready yet
+ *    and needs more input features.
+ *    Note: There is also the function `getScoresWithTimes` which can handle an entire batch of
+ *    requests at once and might be implemented more efficiently (e.g. using batched model forwarding).
+ *  - For all hypotheses that survive pruning, the LabelScorer can compute a new scoring context
+ *    that extends the previous scoring context of that hypothesis with a given successor token. This new
+ *    scoring context can then be used as context in subsequent search steps.
+ *  - After all features have been passed, the `signalNoMoreFeatures` function is called to inform
+ *    the label scorer that it doesn't need to wait for more features and can score as much as possible.
+ *    This is especially important when the label scorer internally uses an encoder or window with right
+ *    context.
+ *  - When all necessary scores for the current segment have been computed, the `reset` function is called
+ *    to clean up any internal data (e.g. feature buffer) or reset flags of the LabelScorer. Afterwards
+ *    it is ready to receive features for the next segment.
+ *
+ * Each concrete subclass internally implements a concrete type of scoring context which the outside
+ * search algorithm is agnostic to. Depending on the model, this scoring context can consist of things like
+ * the current timestep, a label history, a hidden state or other values.
+ */
+class LabelScorer : public virtual Core::Component,
+                    public Core::ReferenceCounted {
+public:
+    // Transition type as part of scoring or context extension requests
+    enum TransitionType {
+        LABEL_TO_LABEL,
+        LABEL_LOOP,
+        LABEL_TO_BLANK,
+        BLANK_TO_LABEL,
+        BLANK_LOOP,
+    };
+
+    // Request for scoring or context extension
+    struct Request {
+        ScoringContextRef context;
+        LabelIndex        nextToken;
+        TransitionType    transitionType;
+    };
+
+    // Return value of scoring function
+    struct ScoreWithTime {
+        Score                  score;
+        Speech::TimeframeIndex timeframe;
+    };
+
+    // Return value of batched scoring function
+    struct ScoresWithTimes {
+        std::vector<Score>                            scores;
+        Core::CollapsedVector<Speech::TimeframeIndex> timesteps;  // Timesteps vector is internally collapsed  if all timesteps are the same (e.g. time-sync decoding)
+    };
+
+    LabelScorer(const Core::Configuration& config);
+    virtual ~LabelScorer() = default;
+
+    // Prepares the LabelScorer to receive new inputs
+    // e.g. by resetting input buffers and segmentEnd flags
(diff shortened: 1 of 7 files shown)


src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -0,0 +1,144 @@
...
+    struct ScoreWithTime {
+        Score                  score;
+        Speech::TimeframeIndex timeframe;
+    };
+
+    // Return value of batched scoring function
+    struct ScoresWithTimes {
+        std::vector<Score>                            scores;
+        Core::CollapsedVector<Speech::TimeframeIndex> timesteps;  // Timesteps vector is internally collapsed  if all timesteps are the same (e.g. time-sync decoding)
+    };
```
larissakl, line 104: Maybe agree on a name here, so either use "timeframe" and "timeframes" or "timestep" and "timesteps".

src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -0,0 +1,144 @@
...
+    // Creates a copy of the context in the request that is extended using the given token and transition type
+    virtual ScoringContextRef extendedScoringContext(Request request) = 0;
+
+    // Function that returns the mapping of each timeframe index (returned in the scoring functions)
+    // to actual flow timestamps with start-/ and end-time in seconds.
+    virtual const std::vector<Flow::Timestamp>& getTimestamps() const = 0;
+
+    // Add a single input feature
+    virtual void addInput(FeatureVectorRef input) = 0;
+    virtual void addInput(Core::Ref<const Speech::Feature> input);
```
larissakl, line 128: Just a little idea: In case one wants to add multiple features at once, for convenience, you could add a method `addInputs()` which takes a batch of features and calls `addInput()` for each.

src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * Each concrete subclass internally implements a concrete type of scoring context which the outside
+ * search algorithm is agnostic to. Depending on the model, this scoring context can consist of things like
+ * the current timestep, a label history, a hidden state or other values.
+ */
+class LabelScorer : public virtual Core::Component,
+                    public Core::ReferenceCounted {
+public:
+    // Transition type as part of scoring or context extension requests
+    enum TransitionType {
```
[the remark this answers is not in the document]
SimBe195: Yes, having this in here is not necessary. I moved it to the search instead.

src/Nn/LabelScorer/LabelScorer.hh
SimBe195: Sure, why not. Done.

src/Nn/LabelScorer/LabelScorer.hh
SimBe195: Fixed.

larissakl (approved):

src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -0,0 +1,140 @@
...
+    struct Request {
+        ScoringContextRef      context;
+        LabelIndex             nextToken;
+        Search::TransitionType transitionType;
+    };
+
+    // Return value of scoring function
+    struct ScoreWithTime {
+        Score                  score;
+        Speech::TimeframeIndex timestep;
```
curufinwe, line 89: The type is called TimeframeIndex, so I would suggest to call this member timeframe instead of timestep.

src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -0,0 +1,140 @@
...
+    };
+
+    // Return value of batched scoring function
+    struct ScoresWithTimes {
+        std::vector<Score>                            scores;
+        Core::CollapsedVector<Speech::TimeframeIndex> timesteps;  // Timesteps vector is internally collapsed  if all timesteps are the same (e.g. time-sync decoding)
```
curufinwe, line 95: see above

src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -0,0 +1,140 @@
...
+
+    // Add a batch of features
+    virtual void addInputs(const std::vector<FeatureVectorRef>& inputs);
+    virtual void addInputs(const std::vector<Core::Ref<const Speech::Feature>>& inputs);
+
+    // Perform scoring computation for a single request
+    // Return score and timeframe index of the corresponding output
+    // May not return a value if the LabelScorer is not ready to score the request yet
+    // (e.g. not enough features received)
+    virtual std::optional<ScoreWithTime> getScoreWithTime(const Request request) = 0;
```
curufinwe, line 130: If the request is passed by value it does not need to be const. But I guess you want to pass by reference here. Also: what do you think about calling this function computeScoreWithTime to better indicate that this is a more costly call?

src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -0,0 +1,140 @@
...
+#include <Search/SearchV2.hh>
+#include <Search/Types.hh>
+#include <Speech/Feature.hh>
+#include <Speech/Types.hh>
+#include <optional>
+#include "ScoringContext.hh"
+
+namespace Nn {
+
+typedef Search::Score                Score;
```
curufinwe, line 38: Do we need to put these into the Nn namespace? Wouldn't it be sufficient to put them in the LabelScorer class?

src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -0,0 +1,140 @@
...
+#include <Core/Types.hh>
+#include <Flow/Timestamp.hh>
+#include <Flow/Vector.hh>
+#include <Mm/FeatureScorer.hh>
+#include <Nn/Types.hh>
+#include <Search/SearchV2.hh>
+#include <Search/Types.hh>
+#include <Speech/Feature.hh>
+#include <Speech/Types.hh>
+#include <optional>
```
curufinwe, line 33: std lib imports should be in a separate block before the RASR imports. Also: newline before importing from the same module (i.e. between Speech/Types.hh and ScoringContext.hh.

src/Nn/LabelScorer/ScoringContext.cc
```diff
@@ -0,0 +1,34 @@
...
+ */
+#include "ScoringContext.hh"
+
+namespace Nn {
+
+typedef Mm::EmissionIndex LabelIndex;
+
+/*
+ * =============================
+ * === ScoringContext ============
```
curufinwe, line 23: too many = chars in this line.

src/Search/SearchV2.hh
```diff
@@ -0,0 +1,30 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+#ifndef SEARCH_V2_HH
+#define SEARCH_V2_HH
+
+namespace Search {
+
+enum TransitionType {
```
curufinwe, line 20: I agree with Tina that this might logically be related to Search, but by putting into the Search namespace we add a dependency for Search in the Nn module. I would like to avoid that as we otherwise might end up having to compile libRasrSearch.<arch>.a whenever we want to use libRasrNn.<arch>.a . So I would prefer to keep it with the LabelScorer in the Nn moudle.

curufinwe (changes requested):
Please write const references as `std::vector<int> const&` instead of `const std::vector<int>&` for better consistency with recently written RASR code.

curufinwe (changes requested):

src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -0,0 +1,145 @@
...
+    virtual void signalNoMoreFeatures() = 0;
+
+    // Gets initial scoring context to use for the hypotheses in the first search step
+    virtual ScoringContextRef getInitialScoringContext() = 0;
+
+    // Creates a copy of the context in the request that is extended using the given token and transition type
+    virtual ScoringContextRef extendedScoringContext(Request const& request) = 0;
+
+    // Add a single input feature
+    virtual void addInput(std::shared_ptr<const f32> const& input, size_t F) = 0;
```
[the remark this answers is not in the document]
SimBe195: In the beginning the input data was buffered using `std::vector`'s which made the logic simple but introduced additional copying-overhead when passing data from Python. So then I changed it to a simple `f32*` but that was problematic because the data could get invalidated after the pointer was pushed to the input buffer of a label scorer. Because of that I changed it to a `shared_ptr<const f32>` which is able to keep the underlying data alive. But `shared_ptr<const f32[]>` seems to work just as well.

There are currently two places where inputs can come from:

1. From a feature flow (i.e. flf recognition). In that case the features are passed as `std::vector`'s which own the data. Then, the label scorer wraps the vector data into a `shared_ptr` that keeps the vector alive in [this way](https://github.com/rwth-i6/rasr/blob/a117a7d6b85a594c5f2c6d537f83876e5bd7633f/src/Nn/LabelScorer/LabelScorer.cc#L29).
2. From a numpy array (i.e. librasr recognition). In that case the data is actually owned externally (in Python) and we pass a `shared_ptr` that contains the `py::array_t<f32>::data()` pointer together with a no-op deleter [like this](https://github.com/rwth-i6/rasr/blob/1576429a217ae561094094ccc94dbff463e8fbee/src/Python/Search.cc#L41).

curufinwe (approved):

curufinwe (approved):

PR #86, opened 2024-12-19
Cleanup old Theano code

larissakl:
As discussed, I removed all (?) Theano-related code which was left, as this is not used anymore. 

In my opinion, it would be a good idea to first merge @SimBe195's CMake compilation system (#71) and remove the Theano Module options in the CMake files as well. This would be helpful for further code cleanups as we could then directly adjust the Makefiles and the CMake files to keep them both clean.


curufinwe (approved):
In principle yes, but if we switch to CMake it would be good to also do it for the AppTek codebase and that has not been tackled yet. That's why I'm holding off on that.

SimBe195 (approved):

PR #85, opened 2024-12-10
Refactor tree construction in AdvancedTreeSearch to allow multiple TreeBuilder types

larissakl:
Integrates the refactoring of the TreeBuilder from AppTek RASR. 
This mainly involves adjustments to allow different tree variants.
Code by @curufinwe

PR #87, opened 2025-01-07
Remove unused support for multiple Trees in SearchSpace construction

larissakl:
We noted that the `trees_` vector in `HMMStateNetwork` is actually superfluous because it always contains at most one element, making it unnecessary to save multiple `Tree` objects. Therefore, I refactored this to use a single member `tree_` instead of a vector.  As a result, constructs such as the `TreeIndex` and the concept of the `masterTree` in `PersistentStateTree` could also be removed.

Since I'm not 100% confident that everything is correct and I want to avoid breaking anything, I have opened this PR to merge into the seq2seq-revamp branch. If you however believe the changes are fine, we could also consider merging it directly into the master branch
This PR builds on the code from PR #85 .

### DIFF
--- a/src/Search/AdvancedTreeSearch/TreeStructure.hh
+++ b/src/Search/AdvancedTreeSearch/TreeStructure.hh
@@ -42,7 +42,7 @@ enum {
     LabelMask                = 1 << 27
 };
 
-///Index of a state or label (see IS_LABEL, ID_FROM_LABEL, and LABEL_FROM_ID)
+/// Index of a state or label (see IS_LABEL, ID_FROM_LABEL, and LABEL_FROM_ID)
 typedef u32 StateId;
 
 ///@todo Maybe this should be zero!
@@ -82,10 +82,10 @@ struct HMMState {
         return *this;
     }
 
-    ///This must be initialized explicitly after creating the state
+    /// This must be initialized explicitly after creating the state
     StateTree::StateDesc stateDesc;
 
-    ///Batch of successor states, managed through a batch-manager in TreeStructure
+    /// Batch of successor states, managed through a batch-manager in TreeStructure
     SuccessorBatchId successors;
 
     /// Returns true if the label-edges batch represents only one single successor, which can be handled more efficiently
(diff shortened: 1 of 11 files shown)


curufinwe:
I merged PR #85 . Could you @larissakl or @SimBe195 update the target branch with it so that we don't see the changes from #85 in this PR? Thanks!

larissakl:
I merged the master branch into seq2seq-revamp, so the code from #85 should be included now

curufinwe:
Hmm, the diff in GitHub still has a lot of the old changes. So something is not right yet.

larissakl:
I see what you mean. Do you have any suggestions how to solve that? I'm not very good at Git, to be honest.

SimBe195:
> Do you have any suggestions how to solve that?

It happens because this was created as a PR into seq2seq-revamp, not into master. Can you change the target of the PR to the master branch @larissakl ?

larissakl:
Sure, but as I said, I'm not completely confident that everything is correct. I tested it with old `AdvancedTreeSearchJob`, but I did not do any further extensive testing.

src/Search/AdvancedTreeSearch/TreeStructure.hh
```diff
@@ -237,68 +228,54 @@ public:
...
-    ///Do not keep pointers to the returned state, the address may change when the network is manipulated
+    /// Do not keep pointers to the returned state, the address may change when the network is manipulated
     inline const HMMState& state(StateId state) const {
         verify_(state > 0 && state < (int)states_.size());
         return states_[state];
     }
 
-    ///Allocates a new tree
-    TreeIndex allocateTree();
+    /// Allocates a new subtree, and adds it into the subtree list.
```
curufinwe, line 243: I don't quite understand the documentation. This could use an update.

curufinwe (changes requested):
Code looks OK, apart from a few small comments. I still need to do some testing.

src/Search/AdvancedTreeSearch/TreeStructure.hh
larissakl: I believe this refers to the "internal handling" of the tree nodes. A new node is not simply appended to the `nodes` vector, this is managed by the `subTreeManager_` which treats each node as a subtree. However I am also not familiar with the details of `Tools::BatchManager` and I agree that this documentation could be clearer. When using this function, the exact implementation details are less important than understanding what it accomplishes. I tried to improve it, please let me know if you have additional suggestions for that.

curufinwe:
I tested your changes with all of the AppTek models and do not see any changes in performance.

curufinwe (approved):

PR #88, opened 2025-01-17
Add BufferedLabelScorer

SimBe195:
Add `BufferedLabelScorer` class.

This is an extension of `LabelScorer` that implements some commonly used buffering logic for input features and timeframes as well as a flag that indicates whether more features are expected to be added to the buffer.

The class serves as a base class for other LabelScorers.


curufinwe (changes requested):
Only one tiny comment.

larissakl (approved):

curufinwe (approved):

PR #89, opened 2025-01-17
Add Encoder class for LabelScorers

SimBe195:
Add `Encoder` class can take features (e.g. from feature flow) and run them through an encoder model to get encoder states. Works with input/output buffer logic, i.e. features get added to an input buffer and outputs are retrieved from an output buffer.

This will later get concrete implementations such as `OnnxEncoder` as subclasses and can be used to pre-process the input features of a LabelScorer in the context of encoder-decoder LabelScorers.

### DIFF
--- a/src/Nn/LabelScorer/Encoder.hh
+++ b/src/Nn/LabelScorer/Encoder.hh
@@ -16,11 +16,12 @@
 #ifndef ENCODER_HH
 #define ENCODER_HH
 
+#include <deque>
+#include <optional>
+
 #include <Core/Component.hh>
 #include <Nn/Types.hh>
 #include <Speech/Feature.hh>
-#include <deque>
-#include <optional>
 
 namespace Nn {
 
@@ -46,7 +47,7 @@ public:
     // Add input features for multiple time steps at once
     virtual void addInputs(std::shared_ptr<const f32[]> const& inputs, size_t timeSize, size_t featureSize);
 
-    // Retrieve a single encoder output
+    // Retrieve the next encoder output frame
     // Performs encoder forwarding internally if necessary
     // Can return None if not enough input features are available yet
     std::optional<std::shared_ptr<const f32[]>> getNextOutput();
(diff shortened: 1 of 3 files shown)


src/Nn/LabelScorer/Encoder.hh
```diff
@@ -0,0 +1,80 @@
...
+    std::optional<std::shared_ptr<const f32[]>> getNextOutput();
+
+    size_t getOutputSize() const;
+
+protected:
+    std::deque<std::shared_ptr<const f32[]>> inputBuffer_;
+    std::deque<std::shared_ptr<const f32[]>> outputBuffer_;
+
+    size_t featureSize_;
+    size_t outputSize_;
```
larissakl, line 62: What's the purpose of `outputSize_`?

src/Nn/LabelScorer/Encoder.hh
```diff
@@ -0,0 +1,80 @@
...
+    bool   expectMoreFeatures_;
+
+    // Consume all features inside input buffer, encode them and put the results into the output buffer
```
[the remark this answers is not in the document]
SimBe195: Good catch, this comment was outdated. I rewrote it now.

src/Nn/LabelScorer/Encoder.hh
SimBe195: It's returned via `getOutputSize` and contains the dimension of the shared_ptr's in the output buffer. This will later be used for example in the `EncoderDecoderLabelScorer` like this:

```
    while ((encoderOutput = encoder_->getNextOutput())) {
        decoder_->addInput(encoderOutput, encoder_->getOutputSize());
    }
```

where `decoder_` is another LabelScorer.
I added a clarifying comment to `getOutputSize`.

larissakl: Ah, I see, thanks for the explanation. Yes, the comment is helpful, otherwise it might not be obvious for someone who is not into the code.

larissakl (approved):

curufinwe (approved):

PR #90, opened 2025-01-22
Add new `SearchAlgorithmV2` interface

SimBe195:
Abstract base class for search algorithms that can work in an online or offline manner.
In contrast to the previous `SearchAlgorithm` this receives features instead of `Mm::FeatureScorers` and the algorithm is supposed to perform scoring internally (usually with a `Nn::LabelScorer`).

The "workflow" usually happens as follows:
 1. Check which `Speech::ModelCombination` (and possibly `Am::AcousticModel`) are needed by the search algorithm via `requiredModelCombination` (and `requiredAcousticModel`).
 2. Create appropriate `Speech::ModelCombination` and provide it to the search algorithm via `setModelCombination`.
 3. Signal segment start via `enterSegment`.
 4. Pass audio features via `passFeature` or `passFeatures`.
 5. Optionally call `decodeStep` or `decodeMore` to run the next search step(s) given the currently available features.
 6. Optionally retrieve intermediate results via `getCurrentBestTraceback` or `getCurrentBestWordLattice`.
 7. Call `finishSegment` to signal that all features have been passed and the search doesn't need to wait for more.
 8. Call `decodeMore` to finalize the search with all the segment features.
 9. Retrieve the final result via `getCurrentBestTraceback` or `getCurrentBestWordLattice`.
 10. Optionally log search statistics via `logStatistics`.
 11. Call `reset` to clean up any buffered features, hypotheses, flags etc. from the previous segment and prepare the algorithm for the next one.
 12. Optionally also reset search statistics via `resetStatistics`.
 13. Continue again at step 3.

Note that the `ScoreVector`, `TracebackItem` and `Traceback` are pretty much copy pasted from old code. Do we want to change anything about them?

### DIFF
--- a/src/Search/SearchV2.hh
+++ b/src/Search/SearchV2.hh
@@ -1,0 +1,163 @@
+/** Copyright 2025 RWTH Aachen University. All rights reserved.
+ *
+ *  Licensed under the RWTH ASR License (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.hltpr.rwth-aachen.de/rwth-asr/rwth-asr-license.html
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+#ifndef SEARCH_V2_HH
+#define SEARCH_V2_HH
+
+#include <Am/AcousticModel.hh>
+#include <Bliss/CorpusDescription.hh>
+#include <Bliss/Lexicon.hh>
+#include <Nn/LabelScorer/LabelScorer.hh>
+#include <Nn/Types.hh>
+#include <Search/LatticeAdaptor.hh>
+#include <Search/Types.hh>
+#include <Speech/Feature.hh>
+#include <Speech/ModelCombination.hh>
+#include <Speech/Types.hh>
+
+namespace Search {
+
+/*
+ * Abstract base class for search algorithms that can work in an online or offline manner.
+ * In contrast to the previous `SearchAlgorithm` this receives features instead of `Mm::FeatureScorers`
+ * and the algorithm is supposed to perform scoring internally (usually with a `Nn::LabelScorer`).
+ *
+ * The "workflow" usually happens as follows:
+ *  1. Check which `Speech::ModelCombination` (and possibly `Am::AcousticModel`) are needed by the search algorithm via
+ *     `requiredModelCombination` (and `requiredAcousticModel`).
+ *  2. Create appropriate `Speech::ModelCombination` and provide it to the search algorithm via `setModelCombination`.
+ *  3. Signal segment start via `enterSegment`.
+ *  4. Add audio features via `addFeature` or `addFeatures`.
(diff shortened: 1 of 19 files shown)


larissakl (approved):
The code looks good to me. 
I believe the `ScoreVector` is fine. Regarding `Traceback` and `TracebackItem`, I’m not entirely sure, but since we’re still using the old `getCurrentBestTraceback()` and `getCurrentBestWordLattice()` functions in the new search algorithms at the moment and we probably won't take the time to rewrite them, I think it’s reasonable to keep them as they are for now.

src/Search/SearchV2.hh
```diff
@@ -0,0 +1,110 @@
...
+    virtual void reset() = 0;
+
+    // Signal the beginning of a new audio segment.
+    virtual void enterSegment(Bliss::SpeechSegment const* = nullptr) = 0;
+
+    // Signal that all features of the current segment have been passed.
+    virtual void finishSegment() = 0;
+
+    // Pass a single feature vector.
+    virtual void passFeature(std::shared_ptr<const f32[]> const& data, size_t featureSize) = 0;
```
curufinwe, line 81: How about `putFeature` instead of `passFeature`?

src/Search/SearchV2.hh
```diff
@@ -0,0 +1,110 @@
...
+    virtual Core::Ref<const Traceback> getCurrentBestTraceback() const = 0;
+
+    // Similar to `getCurrentBestTraceback` but return the lattice instead of just single-best traceback.
+    virtual Core::Ref<const LatticeAdaptor> getCurrentBestWordLattice() const = 0;
+
+    // Try to decode one more step. Return bool indicates whether a step could be made.
+    virtual bool decodeStep() = 0;
+
+    // Decode as much as possible given the currently available features. Return bool indicates whether any steps could be made.
+    virtual bool decodeMore() {
```
curufinwe, line 97: How about `decodeManySteps` instead of `decodeMore`?

src/Search/SearchV2.hh
```diff
@@ -0,0 +1,110 @@
...
+    virtual Core::Ref<const Traceback> getCurrentBestTraceback() const = 0;
+
+    // Similar to `getCurrentBestTraceback` but return the lattice instead of just single-best traceback.
+    virtual Core::Ref<const LatticeAdaptor> getCurrentBestWordLattice() const = 0;
+
+    // Try to decode one more step. Return bool indicates whether a step could be made.
+    virtual bool decodeStep() = 0;
+
+    // Decode as much as possible given the currently available features. Return bool indicates whether any steps could be made.
+    virtual bool decodeManySteps() {
```
larissakl, line 97: I would personally prefer something like `runDecodeSteps` because I think "many" is misleading here, it could be only one step.

curufinwe: Hmm, I would definitely not add `run` as a prefix as that's not adding more information. On the ambiguity whether it's one step or many: It can also be zero here as well as above, so that is already part of the API. How about we return the number of executed steps here instead of a boolean flag? The returned unsigned integer can be 0, 1 or more.

larissakl: I'm not sure, do you think this information about how many steps have been executed would be useful at some point? If not, I would keep the boolean flag.

curufinwe: It does not hurt to have it here. If you want to check for success you can still do it with an integer as return value. But I would mostly add it to indicate that a variable number of steps can be executed.

curufinwe (approved):

PR #91, opened 2025-01-27
`LexiconfreeBeamSearch`: Refactor decoding step and add length normalization

larissakl:
To make the decoding step of the `LexiconfreeBeamSearch` clearer and more readable, I refactored the code by separating the different pruning steps into dedicated helper functions. 
Additionally, I introduced the option to normalize the scores based on the hypothesis length. The normalization factor can be scaled with the parameter `length-norm-scale`.

### DIFF
--- a/src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.cc
+++ b/src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.cc
@@ -231,6 +231,57 @@ std::string LexiconfreeBeamSearch::LabelHypothesis::toString() const {
     return ss.str();
 }
 
+void LexiconfreeBeamSearch::tokenPruning(std::vector<size_t>& indices, std::vector<Score> extensionScores, size_t numUnfinishedHyps) {
+    #pragma omp parallel for
+        for (size_t hypIndex = 0ul; hypIndex < numUnfinishedHyps; ++hypIndex) {
+            // Start and end index of scores coming from current hypothesis
+            size_t start = hypIndex * lexicon_->nLemmas();
+            size_t end   = start + lexicon_->nLemmas();
+
+            // Make sure the best topKTokens_ scores are moved to the beginning of the [start, end) interval
+            std::nth_element(indices.begin() + start, indices.begin() + start + topKTokens_, indices.begin() + end,
+                             [&extensionScores](size_t a, size_t b) {
+                                 return extensionScores[a] < extensionScores[b];
+                             });
+
+            // Copy the topKTokens_ best indices to the front of indices (note: this location doesn't overlap with [start, end) of the next hyps because topKTokens_ < nLemmas)
+            if (hypIndex > 0ul) {
+                std::copy(indices.begin() + start, indices.begin() + start + topKTokens_, indices.begin() + hypIndex * topKTokens_);
+            }
+        }
+
+    indices.resize(numUnfinishedHyps * topKTokens_);  // Throw away all pruned indices
+}
+
+template <typename Hypotheses, typename CompareFunc>
+void LexiconfreeBeamSearch::beamPruning(Hypotheses& hypotheses, CompareFunc compare) {
+    if (hypotheses.size() > maxBeamSize_) {
+        // Sort the hypotheses by associated score value such that the first `beamSize_` elements are the best and sorted
+        std::partial_sort(hypotheses.begin(), hypotheses.begin() + maxBeamSize_, hypotheses.end(), compare);
+        hypotheses.resize(maxBeamSize_);  // Get rid of excessive elements
+    }
+    else {
+        // Fully sort the hypotheses by their score
+        std::sort(hypotheses.begin(), hypotheses.end(), compare);
+    }
+}
+
+template <typename Hypotheses, typename GetScoreFunc>
+void LexiconfreeBeamSearch::scorePruning(Hypotheses& hypotheses, GetScoreFunc getScore) {
+    // Compute the pruning threshold
+    auto   pruningThreshold    = getScore(hypotheses.front()) + scoreThreshold_;
+    size_t numSurvivingHyps = 0ul;
+    // Use the fact that hypotheses are sorted by corresponding score and prune all indices after the first one that
+    // violates t&he score threshold
+    for (auto& hyp : hypotheses) {
+        if (getScore(hyp) > pruningThreshold) {
+            break;
+        }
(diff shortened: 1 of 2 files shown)


src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.cc
```diff
@@ -287,11 +341,19 @@ bool LexiconfreeBeamSearch::decodeStep() {
...
-    auto const& scoresWithTimes = result.value();
+    auto const& scoresWithTimes        = result.value();
+    std::vector<Score> extensionScores = scoresWithTimes.scores;
 
     std::vector<float> combinedScores(numUnfinishedHyps * nLemmas);
     for (size_t requestIndex = 0ul; requestIndex < requests.size(); ++requestIndex) {
-        combinedScores[requestIndex] = baseHyps[requestIndex]->score + scoresWithTimes.scores[requestIndex];
+        if (useLengthNormalization_) {
+            // With length normalization: combined score = (current score * (current length * lengthNormScale_) + score of this extension) / ((current length + 1) * lengthNormScale_ )
+            combinedScores[requestIndex] = (baseHyps[requestIndex]->score * (baseHyps[requestIndex]->length * lengthNormScale_) + extensionScores[requestIndex]) / ((baseHyps[requestIndex]->length + 1) * lengthNormScale_);
```
SimBe195, line 351: The `lengthNormScale_` should be an exponent, not a multiplier. The formula is $\text{score}_\text{norm} = \frac{\text{score}}{\text{length}^\alpha}$ so that the default value of $\alpha = 0$ means no normalization.

src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.cc
```diff
@@ -228,6 +231,57 @@ std::string LexiconfreeBeamSearch::LabelHypothesis::toString() const {
     return ss.str();
 }
 
+void LexiconfreeBeamSearch::tokenPruning(std::vector<size_t>& indices, std::vector<Score> extensionScores, size_t numUnfinishedHyps) {
```
SimBe195, line 234: Rename to `topKTokenPruning` to make it more explicit.

src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.cc
```diff
@@ -228,6 +231,57 @@ std::string LexiconfreeBeamSearch::LabelHypothesis::toString() const {
...
+            if (hypIndex > 0ul) {
+                std::copy(indices.begin() + start, indices.begin() + start + topKTokens_, indices.begin() + hypIndex * topKTokens_);
+            }
+        }
+
+    indices.resize(numUnfinishedHyps * topKTokens_);  // Throw away all pruned indices
+}
+
+template <typename Hypotheses, typename CompareFunc>
+void LexiconfreeBeamSearch::beamPruning(Hypotheses& hypotheses, CompareFunc compare) {
```
SimBe195, line 257:
```suggestion
template <typename T>
void LexiconfreeBeamSearch::beamPruning(std::vector<T>& hypotheses, std::function<bool,(T const&, T const&)>&& compare) {
```
You can rewrite the template definition to make it clearer how `compare` is supposed to look like.

src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.cc
```diff
@@ -228,6 +231,57 @@ std::string LexiconfreeBeamSearch::LabelHypothesis::toString() const {
...
+        hypotheses.resize(maxBeamSize_);  // Get rid of excessive elements
+    }
+    else {
+        // Fully sort the hypotheses by their score
+        std::sort(hypotheses.begin(), hypotheses.end(), compare);
+    }
+}
+
+template <typename Hypotheses, typename GetScoreFunc>
+void LexiconfreeBeamSearch::scorePruning(Hypotheses& hypotheses, GetScoreFunc getScore) {
```
SimBe195, line 270: See above.

src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.hh
```diff
@@ -23,33 +23,36 @@
...
 // Given a lexicon only containing labels (without lemmas), pick the label index with
 // maximum probability at each decoding step.
+// Supports pruning of the top-k successor of each hypothesis, max-beam-size-pruning and score-based pruning.
 class LexiconfreeBeamSearch : public SearchAlgorithmV2 {
     struct HypothesisExtension {
         const Bliss::Lemma*             lemma;
         Nn::ScoringContextRef           scoringContext;
         Nn::LabelIndex                  label;
         Score                           score;
+        int                             length;
```
SimBe195, line 36: `HypothesisExtension` doesn't need to contain a `length` member.

src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.hh
```diff
@@ -23,33 +23,36 @@
...
-                : lemma(lemma), scoringContext(scoringContext), label(label), score(score), timestep(timestep), transitionType(transitionType) {}
+        HypothesisExtension(const Bliss::Lemma* lemma, Core::Ref<const Nn::ScoringContext> scoringContext, Nn::LabelIndex label, Score score, int length, Search::TimeframeIndex timestep, Nn::LabelScorer::TransitionType transitionType)
+                : lemma(lemma), scoringContext(scoringContext), label(label), score(score), length(length), timestep(timestep), transitionType(transitionType) {}
     };
 
     struct LabelHypothesis {
         Nn::ScoringContextRef scoringContext;
         Nn::LabelIndex        currentLabel;
         Score                 score;
+        int                   length;
```
SimBe195, line 51: Length can't be negative so it should be unsigned.

src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.hh
```diff
@@ -118,6 +139,9 @@ private:
     bool  useScorePruning_;
     Score scoreThreshold_;
 
+    bool useLengthNormalization_;
+    float lengthNormScale_;
```
SimBe195, line 143: Prefer to use the types defined by RASR (e.g. `f32` or `Score` in this case).

SimBe195 (changes requested):

PR #93, opened 2025-01-28
Add CtcTreeBuilder

larissakl:
Adds a `CtcTreeBuilder` to build a search tree that respects the CTC topology.
Each node has a self-loop to allow for label repetitions. Between two labels, there is always a "blank" label node, which is however optional due to skip-transitions. These skip-transitions are however omitted for repeating labels.
If a word-boundary token exists in the lexicon, a word-boundary-root will be added and all word ends (except for blank and silence) will be redirected to this root. The transition state of the word-boundary token itself is the "normal" root, such that a new word can be started afterward.

Until now, there is no support for context-dependent labels.
In a next step, we need to move the tree building out of the `AdvancedTreeSearch`, such that it can easily be used by other search algorithms.

This code was mostly written by @curufinwe, I added some modifications.
This PR is dependent on #87.


src/Search/AdvancedTreeSearch/SearchSpace.cc
```diff
@@ -499,9 +499,9 @@ void StaticSearchAutomaton::clearDepths() {
 
 int StaticSearchAutomaton::fillStateDepths(StateId state, int depth) {
     if (stateDepths[state] != Core::Type<int>::min) {
-        if (stateDepths[state] != depth) {  /// @todo Find out why this happens on some languages
-            std::cout << "conflicting state depths: " << stateDepths[state] << " vs " << depth << std::endl;
-        }
+        //if (stateDepths[state] != depth) {  /// @todo Find out why this happens on some languages
+        //    std::cout << "conflicting state depths: " << stateDepths[state] << " vs " << depth << std::endl;
+        //}
```
curufinwe, line 504:
```suggestion
```

src/Search/AdvancedTreeSearch/PersistentStateTree.hh
```diff
@@ -100,6 +104,9 @@ public:
     // Context-independent root node
     StateId ciRootState;
 
+    // Other root nodes (currently used for the wordBoundaryRoot in CtcTreeBuilder)
+    std::vector<StateId> otherRootStates;
```
curufinwe, line 108: Why not a std::set?

curufinwe (changes requested):
Only small changes.

src/Search/AdvancedTreeSearch/SearchSpace.cc
larissakl: So I commented that out as it was triggered with the CtcTreeBuilder because of the self-loops. However, I don't think we would actually use the CtcTreeBuilder with the AdvancedTreeSearch, I could actually uncomment it again. Or do you want to remove it nevertheless?

curufinwe: If CTC breaks this check then the check is no longer needed. I also would like to try using CTC with AdvancedTreeSearch. This could be a useful comparison for any new thing we are building.

larissakl: Ok, done.

curufinwe (approved):

src/Search/AdvancedTreeSearch/TreeBuilder.cc
```diff
@@ -1197,3 +1197,200 @@ inline void MinimizedTreeBuilder::mapSuccessors(const std::set<StateId>& success
...
+                currentState = extendState(currentState, desc);
+                // Add loop for this state
+                addTransition(currentState, currentState);
+                // Add transition from previous non-blank state to this state, allowing to skip the blank state in-between these two
+                if (prevNonBlankState != invalidTreeNodeIndex) {
+                    addTransition(prevNonBlankState, currentState);
+                }
+                prevNonBlankState = currentState;
+
+                bool is_last_state_in_lemma = ((phoneState + 1) == hmmTopology->nPhoneStates()) and ((subState + 1) == hmmTopology->nSubStates()) and (boundary & Am::Allophone::isFinalPhone);
```
SimBe195, line 1352: `is_last_state_in_lemma` -> `isLastStateInLemma`

src/Search/AdvancedTreeSearch/TreeBuilder.cc
```diff
@@ -1197,3 +1197,200 @@ inline void MinimizedTreeBuilder::mapSuccessors(const std::set<StateId>& success
...
+
+        Bliss::ContextPhonology::SemiContext history, future;
+        const Am::Allophone*                 allophone          = acousticModel_.allophoneAlphabet()->allophone(Am::Allophone(Bliss::ContextPhonology::PhonemeInContext(phoneme, history, future), boundary));
+        const Am::ClassicHmmTopology*        hmmTopology        = acousticModel_.hmmTopology(phoneme);
+        const bool                           allophone_is_blank = acousticModel_.allophoneStateAlphabet()->index(allophone, 0, false) == blankAllophoneStateIndex_;
+
+        for (u32 phoneState = 0; phoneState < hmmTopology->nPhoneStates(); ++phoneState) {
+            Am::AllophoneState   alloState = acousticModel_.allophoneStateAlphabet()->allophoneState(allophone, phoneState);
+            StateTree::StateDesc desc;
+            desc.acousticModel = acousticModel_.emissionIndex(alloState);  // Decision tree look-up for CART id.
```
SimBe195, line 1336:
```suggestion
            desc.acousticModel = acousticModel_.emissionIndex(alloState);  // state-tying look-up
```

SimBe195 (changes requested):

PR #94, opened 2025-01-28
Add OnnxEncoder + EncoderFactory

SimBe195:
Add a class `OnnxEncoder` which inherits from `Nn::Encoder` and works by forwarding its inputs through an Onnx model.

This PR also contains a `EncoderFactory` in `Nn::Module` which is used for registering and creating encoder instances (this could also be done in a separate PR if needed).


src/Onnx/Value.hh
```diff
@@ -58,6 +58,12 @@ public:
     template<typename T>
     static Value zeros(std::vector<int64_t> const& dim);
 
+    template<typename T>
```
curufinwe, line 61: This is a subset of the zeros constructor. I would put these 2 functions before zeros and use their implementation inside of `zeros`.

src/Onnx/OnnxEncoder.cc
```diff
@@ -0,0 +1,111 @@
...
+void OnnxEncoder::encode() {
+    if (inputBuffer_.empty()) {
+        return;
+    }
+
+    /*
+     * Create session inputs
+     */
+
+    std::vector<std::pair<std::string, Value>> sessionInputs;
```
curufinwe, line 61:
```suggestion
     */
    std::vector<std::pair<std::string, Value>> sessionInputs;
```

src/Onnx/OnnxEncoder.cc
```diff
@@ -0,0 +1,111 @@
...
+    // features-size is an optional input
+    if (featuresSizeName_ != "") {
+        sessionInputs.emplace_back(std::make_pair(featuresSizeName_, Value::create(std::vector<s32>{static_cast<int>(T_in)})));
+    }
+
+    /*
+     * Run session
+     */
+
+    std::vector<Value> sessionOutputs;
```
curufinwe, line 84:
```suggestion
     */
    std::vector<Value> sessionOutputs;
```

src/Onnx/OnnxEncoder.cc
```diff
@@ -0,0 +1,111 @@
...
+    onnxModel_.session.run(std::move(sessionInputs), {outputName_}, sessionOutputs);
+
+    /*
+     * Put outputs into buffer
+     */
+
+    auto onnxOutputValueWrapper = std::make_shared<Value>(std::move(sessionOutputs.front()));
```
curufinwe, line 91:
```suggestion
     */
    auto onnxOutputValueWrapper = std::make_shared<Value>(std::move(sessionOutputs.front()));
```

src/Onnx/OnnxEncoder.cc
```diff
@@ -0,0 +1,111 @@
...
+
+    size_t T_out;
+    if (onnxOutputValueWrapper->numDims() == 3) {
+        T_out       = onnxOutputValueWrapper->dimSize(1);
+        outputSize_ = onnxOutputValueWrapper->dimSize(2);
+    }
+    else {
```
curufinwe, line 98: If 2-dimensional outputs are supported we should add that to the io spec above.

src/Onnx/OnnxEncoder.cc
curufinwe, line 91: Not sure if we do it anywhere else, but we should get the index of the output named outputName_ here and not assume it's the first one.

curufinwe (changes requested):

larissakl (approved):

src/Onnx/OnnxEncoder.cc
SimBe195: @curufinwe `session.run` doesn't compute all possible outputs in the ONNX model but only the outputs requested via the `output_names` parameter which is set to `{outputName_}`. So `sessionOutputs` is always a size-1-vector with the single element corresponding to the `outputName_` output.

curufinwe: OK

curufinwe (approved):

PR #95, opened 2025-02-06
Add LabelScorerFactory

SimBe195:
Factory class to register types of LabelScorers and create them.

Introduced so that LabelScorers can be registered from different places in the codebase (e.g. inside src/Nn/LabelScorer and src/Onnx).

The `Nn::Module_` instance keeps a `labelScorerFactory_` member through which the factory can be accessed.

PR #97, opened 2025-02-06
Add CombineLabelScorer

SimBe195:
Add `CombineLabelScorer` which is a LabelScorer that performs log-linear combination of multiple sub-label-scorers, assuming that the sub-scorers have the same label alphabet,
i.e. $$\text{combined\\_score}(\text{request}) = \sum_k \text{score}_k(\text{request}) * \text{scale}_k$$

The assigned timeframe is the maximum over the sub-timeframes,
i.e. $$\text{combined\\_timeframe}(\text{request}) = \max_k \lbrace \text{timeframe}_k(\text{request}) \rbrace$$.

Depends on #95.


larissakl:
I have a general question, maybe I am missing something, but what's the purpose to `ScoringContextHash` and the `hash()` function of the different ScoringContexts?

SimBe195:
> I have a general question, maybe I am missing something, but what's the purpose of `ScoringContextHash` and the `hash()` function of the different ScoringContexts?

@larissakl The hash (and equality) is needed to be able to put ScoringContext objects into datastructures like hashmaps or hashsets. For example here is a usage:
[https://github.com/rwth-i6/rasr/blob/bd6333e839aadddc9f3e14f236fa361424c86e5b/src/Nn/LabelScorer/LimitedCtxOnnxLabelScorer.hh#L91](https://github.com/rwth-i6/rasr/blob/bd6333e839aadddc9f3e14f236fa361424c86e5b/src/Nn/LabelScorer/LimitedCtxOnnxLabelScorer.hh#L91)
or here
[https://github.com/rwth-i6/rasr/blob/bd6333e839aadddc9f3e14f236fa361424c86e5b/src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.cc#L270](https://github.com/rwth-i6/rasr/blob/bd6333e839aadddc9f3e14f236fa361424c86e5b/src/Search/LexiconfreeBeamSearch/LexiconfreeBeamSearch.cc#L270)

The operations are implemented as virtual member functions of `ScoringContext` which are called inside `ScoringContextHash` and `ScoringContextEq` so that each subclass can override it with custom implementations.

curufinwe (approved):

larissakl:
Ah, I see, I didn't think of that. Thank you @SimBe195 .

larissakl (approved):

PR #98, opened 2025-02-11
Move tree construction related code from AdvancedTreeSearch into Search

larissakl:
To improve modularity and make the tree building easily reusable for new search algorithms, I moved all tree-building-related code into a dedicated subdirectory (Search/TreeBuilder).
Previously, `StaticSearchAutomaton` contained the function `createTreeBuilder()`, which instantiated a tree builder based on the selected `TreeBuilderType`. I moved this function to TreeBuilder.hh, so that it can be used more flexibly in other parts.
This is my proposed approach, I’m open to feedback and discussions, maybe you have alternative ideas or suggestions.

Depends on #93 , as it already includes the `CtcTreeBuilder`.


SimBe195 (approved):

larissakl:
Update:

As discussed, I removed the TreeBuilder subdirectory, moved the tree builder code into Search and relocated the function `createTreeBuilder()` to `Search::Module` instead of `TreeBuilder.hh`.
Let me know if I missed anything or if you had different expectations.
There are two points I’m unsure about:
1. Would you prefer to move `choiceTreeBuilderType` and `paramTreeBuilderType` to `Search::Module` as well or should we keep them in `SearchSpace` (and in all future search algorithms that utilize a search tree)?
2. For `TreeBuilderType::classicHmm`, `createTreeBuilder() `currently returns `std::unique_ptr<AbstractTreeBuilder>(nullptr)`, meaning a null pointer must be handled explicitly. Is this acceptable for you or would you prefer a different approach?

curufinwe:
1. Move BatchManager into Search
2. Make createTreeBuilder only take a config and have the choice parameter inside the _Module class.

curufinwe:
I removed some more dependencies inside of Search to AdvancedTreeSearch and then also unified the include guards and include order in these files + ran clang-format. That's why this PR got a bit noisy. I'll add a separate PR to add a github action that checks for code-formatting with clang-format version 14 so we don't need to have these clang-format discussions anymore.

curufinwe (approved):

SimBe195 (approved):

PR #99, opened 2025-02-15
Add TransducerTreeBuilder

larissakl:
This PR introduces a `TransducerTreeBuilder` for constructing a search tree with Transducer topology, similar to #93.
Since it shares most of its logic, I made it a subclass of `CtcTreeBuilder`. The only difference is the absence of self-loops on the (non-blank) labels.
Depends on #98.

### DIFF
--- a/src/Search/AdvancedTreeSearch/SearchSpace.cc
+++ b/src/Search/AdvancedTreeSearch/SearchSpace.cc
@@ -136,6 +136,7 @@ const Core::Choice choiceTreeBuilderType(
         "classic-hmm", static_cast<int>(TreeBuilderType::classicHmm),
         "minimized-hmm", static_cast<int>(TreeBuilderType::minimizedHmm),
         "ctc", static_cast<int>(TreeBuilderType::ctc),
+        "transducer", static_cast<int>(TreeBuilderType::transducer),
         Core::Choice::endMark());
 
 const Core::ParameterChoice paramTreeBuilderType(
--- a/src/Search/TreeBuilder/TreeBuilder.cc
+++ b/src/Search/TreeBuilder/TreeBuilder.cc
@@ -1395,4 +1395,64 @@ void CtcTreeBuilder::addWordBoundaryStates() {
     addTransition(blankBefore, blankBefore);
 }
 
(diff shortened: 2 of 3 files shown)


SimBe195:
Since the only difference between `CtcTreeBuilder` and `TransducerTreeBuilder` is leaving out this part in `extendPronunciation`
````
                // Add loop for this state
                addTransition(currentState, currentState);
````
I would probably prefer to unify them by adding a member `bool labelLoop_` to the `CtcTreeBuilder` which can be set via the constructor (defaulting to `true`). Then in `createTreeBuilder` you would do
````
        case TreeBuilderType::ctc: {
            return std::unique_ptr<AbstractTreeBuilder>(new CtcTreeBuilder(config, lexicon, acousticModel, network, initialize, true));
        } break;
        case TreeBuilderType::transducer: {
            return std::unique_ptr<AbstractTreeBuilder>(new CtcTreeBuilder(config, lexicon, acousticModel, network, initialize, false));
        } break;
````
Alternatively, `allow-label-loop` could also be a config parameter.

larissakl:
@SimBe195 Yes, I actually had the same idea. Should we, however, reconsider the naming of `CtcTreeBuilder`? From a user's perspective, it's not a problem, but maybe it's too unintuitive for people working with this code in the future?

SimBe195:
> @SimBe195 Yes, I actually had the same idea. Should we, however, reconsider the naming of `CtcTreeBuilder`? From a user's perspective, it's not a problem, but maybe it's too unintuitive for people working with this code in the future?

I think `CtcTreeBuilder` is fine. Perhaps `"transducer"` could be changed to `"rna"` though to be consistent with the naming that was chosen in the `AllophoneStateGraphBuilder` (see [here](https://github.com/rwth-i6/rasr/blob/21b9645a91d70e4c440e2924377d024dbb9ec90c/src/Speech/Module.cc#L128)). From a user perspective it's probably easiest to just choose a topology by name and not have to set config parameters to enable/disable label loops.

larissakl:
Yes, I definitely prefer choosing the topology by name.
Ok, I will just close this PR, implement the changes and open a new one.

PR #100, opened 2025-02-18
Add RnaTreeBuilder

larissakl:
As discussed in #99, the difference between the `CtcTreeBuilder` and a tree builder for Transducer/RNA topology is minimal and a dedicated `TransducerTreeBuilder` subclass can be avoided.
Therefore I introduced a member `labelLoops_` to `CtcTreeBuilder`, which determines whether self-loops on the non-blank labels are added (CTC topology) or not (RNA topology).
Depends on #98.


SimBe195 (approved):

larissakl:
Update:

I introduced the `RnaTreeBuilder` class as a subclass of `CtcTreeBuilder`. In `SearchSpace.cc`, there is now a new config-parameter, `allow-label-loop`.
- If `allow-label-loop` is not set, it defaults to true for `CtcTreeBuilder` ("classic" CTC topology) and false for `RnaTreeBuilder` ("classic" RNA topology).
- If `tree-builder-type = rna` and `allow-label-loop = yes`, the `RnaTreeBuilder` class is still used, but it will effectively build a tree with CTC topology. 
- Conversely, if `tree-builder-type = ctc` and `allow-label-loop = no`, the `CtcTreeBuilder` will build a tree with RNA topology.

I’m still unsure about the code style and the intended behavior, so feel free to share feedback or make any necessary changes directly.

curufinwe (approved):

PR #101, opened 2025-02-19
Lexiconfree beam search

SimBe195:
Simple time synchronous beam search algorithm based on the new `SearchAlgorithmV2` interface. Does not use (proper) pronunciation lexicon, word-level LM or transition model. Performs special handling of blank if a blank index is set. Main purpose is open vocabulary search with CTC/Neural Transducer (or similar) models.

Supports global pruning by max beam-size and by score difference to the best hypothesis. Uses a LabelScorer to context initialization/extension and scoring.

The search requires a lexicon that represents the vocabulary. Each lemma is viewed as a token with its index in the lexicon corresponding to the associated output index of the LabelScorer.

### DIFF
--- a/src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
+++ b/src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
@@ -1,0 +1,442 @@
+/** Copyright 2025 RWTH Aachen University. All rights reserved.
+ *
+ *  Licensed under the RWTH ASR License (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.hltpr.rwth-aachen.de/rwth-asr/rwth-asr-license.html
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#include "LexiconfreeTimesyncBeamSearch.hh"
+#include <Core/XmlStream.hh>
+#include <Lattice/LatticeAdaptor.hh>
+#include <Nn/LabelScorer/LabelScorer.hh>
+#include <Nn/LabelScorer/ScoringContext.hh>
+#include <algorithm>
+#include <strings.h>
+
+namespace Search {
+
+/*
+ * =====================================
+ * === LexiconfreeTimesyncBeamSearch ===
+ * =====================================
+ */
+
+const Core::ParameterInt LexiconfreeTimesyncBeamSearch::paramMaxBeamSize(
+        "max-beam-size",
+        "Maximum number of elements in the search beam.",
+        1, 1);
+
+const Core::ParameterFloat LexiconfreeTimesyncBeamSearch::paramScoreThreshold(
+        "score-threshold",
+        "Prune any hypotheses with a score that is at least this much worse than the best hypothesis. If not set, no score pruning will be done.",
+        Core::Type<Score>::max, 0);
+
+const Core::ParameterInt LexiconfreeTimesyncBeamSearch::paramBlankLabelIndex(
+        "blank-label-index",
+        "Index of the blank label in the lexicon. If not set, the search will not use blank.",
+        Core::Type<int>::max);
+
+const Core::ParameterBool LexiconfreeTimesyncBeamSearch::paramAllowLabelLoop(
+        "allow-label-loop",
+        "Collapse repeated emission of the same label into one output. If false, every emission is treated like a new output.",
+        false);
+
+const Core::ParameterBool LexiconfreeTimesyncBeamSearch::paramLogStepwiseStatistics(
+        "log-stepwise-statistics",
+        "Log statistics about the beam at every search step.",
+        false);
+
+const Core::ParameterBool LexiconfreeTimesyncBeamSearch::paramDebugLogging(
+        "debug-logging",
+        "Enable detailed logging for debugging purposes.",
+        false);
+
+LexiconfreeTimesyncBeamSearch::LexiconfreeTimesyncBeamSearch(Core::Configuration const& config)
+        : Core::Component(config),
+          SearchAlgorithmV2(config),
+          maxBeamSize_(paramMaxBeamSize(config)),
+          scoreThreshold_(paramScoreThreshold(config)),
+          blankLabelIndex_(paramBlankLabelIndex(config)),
+          allowLabelLoop_(paramAllowLabelLoop(config)),
+          logStepwiseStatistics_(paramLogStepwiseStatistics(config)),
+          debugLogging_(paramDebugLogging(config)),
+          labelScorer_(),
+          beam_(),
+          initializationTime_(),
+          featureProcessingTime_(),
+          scoringTime_(),
+          contextExtensionTime_() {
+    beam_.reserve(maxBeamSize_);
+    useBlank_        = blankLabelIndex_ != Core::Type<int>::max;
+    useScorePruning_ = scoreThreshold_ != Core::Type<Score>::max;
+}
+
+void LexiconfreeTimesyncBeamSearch::reset() {
+    initializationTime_.tic();
+
+    labelScorer_->reset();
+
+    // Reset beam to a single empty hypothesis
+    beam_.clear();
+    beam_.push_back(LabelHypothesis());
+    beam_.front().scoringContext = labelScorer_->getInitialScoringContext();
+
+    initializationTime_.toc();
+}
+
+Speech::ModelCombination::Mode LexiconfreeTimesyncBeamSearch::requiredModelCombination() const {
+    return Speech::ModelCombination::useLabelScorer | Speech::ModelCombination::useLexicon;
+}
+
+bool LexiconfreeTimesyncBeamSearch::setModelCombination(Speech::ModelCombination const& modelCombination) {
+    lexicon_     = modelCombination.lexicon();
+    labelScorer_ = modelCombination.labelScorer();
+
+    reset();
+    return true;
+}
+
+void LexiconfreeTimesyncBeamSearch::enterSegment(Bliss::SpeechSegment const* segment) {
+    initializationTime_.tic();
+    labelScorer_->reset();
+    resetStatistics();
+    initializationTime_.toc();
+}
+
+void LexiconfreeTimesyncBeamSearch::finishSegment() {
+    featureProcessingTime_.tic();
+    labelScorer_->signalNoMoreFeatures();
+    featureProcessingTime_.toc();
+    decodeManySteps();
+    logStatistics();
+}
+
+void LexiconfreeTimesyncBeamSearch::putFeature(std::shared_ptr<const f32[]> const& data, size_t featureSize) {
+    featureProcessingTime_.tic();
+    labelScorer_->addInput(data, featureSize);
+    featureProcessingTime_.toc();
+}
+
+void LexiconfreeTimesyncBeamSearch::putFeature(std::vector<f32> const& data) {
+    featureProcessingTime_.tic();
+    labelScorer_->addInput(data);
+    featureProcessingTime_.toc();
+}
+
+void LexiconfreeTimesyncBeamSearch::putFeatures(std::shared_ptr<const f32[]> const& data, size_t timeSize, size_t featureSize) {
+    featureProcessingTime_.tic();
+    labelScorer_->addInputs(data, timeSize, featureSize);
+    featureProcessingTime_.toc();
+}
+
+Core::Ref<const Traceback> LexiconfreeTimesyncBeamSearch::getCurrentBestTraceback() const {
+    return Core::ref(new Traceback(beam_.front().traceback));
+}
+
+Core::Ref<const LatticeAdaptor> LexiconfreeTimesyncBeamSearch::getCurrentBestWordLattice() const {
+    if (beam_.front().traceback.empty()) {
+        return Core::ref(new Lattice::WordLatticeAdaptor());
+    }
+
+    // use default LemmaAlphabet mode of StandardWordLattice
+    Core::Ref<Lattice::StandardWordLattice> result(new Lattice::StandardWordLattice(lexicon_));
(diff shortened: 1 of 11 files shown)


larissakl:
Two general points:
1. How should this search algorithm be used? Are you planning to put the `enum SearchTypeV2`, the related `ParameterChoice` and `Module_::createSearchAlgorithm()` to Module.hh/.cc later? Or do you have a different plan?
2. Would it maybe be a good idea to factor out the `struct TimeStatistic` and its related code so that it can be reused in other search algorithms?

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.hh
```diff
@@ -0,0 +1,174 @@
...
+     * Struct containing all information about a single hypothesis in the beam
+     */
+    struct LabelHypothesis {
+        Nn::ScoringContextRef scoringContext;  // Context to compute scores based on this hypothesis
+        Nn::LabelIndex        currentToken;    // Most recent token in associated label sequence (useful to infer transition type)
+        Score                 score;           // Full score of hypothesis
+        Traceback             traceback;       // Associated traceback to return
+
+        LabelHypothesis();
+        LabelHypothesis(LabelHypothesis const& base);
```
larissakl, line 66: I think this constructor is not implemented and not needed anymore

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,444 @@
...
+
+const Core::ParameterFloat LexiconfreeTimesyncBeamSearch::paramScoreThreshold(
+        "score-threshold",
+        "Prune any hypotheses with a score that is at least this much worse than the best hypothesis. If not set, no score pruning will be done.",
+        Core::Type<Score>::max, 0);
+
+const Core::ParameterInt LexiconfreeTimesyncBeamSearch::paramBlankLabelIndex(
+        "blank-label-index",
+        "Index of the blank label in the lexicon. If not set, the search will not use blank.",
+        Core::Type<int>::max);
```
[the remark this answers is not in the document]
SimBe195: I have now made it so that the index can be set both via config parameter or via special lemma. If the two values disagree, the search uses the value from the config and emits a warning.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,444 @@
...
+
+    /*
+     * Prune set of possible extensions by max beam size and possibly also by score.
+     */
+    beamPruning(extensions);
+    if (debugLogging_) {
+        log() << extensions.size() << " candidates survived beam pruning";
+    }
+
+    std::sort(extensions.begin(), extensions.end());
```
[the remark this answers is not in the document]
SimBe195: Yes, the recombination also requires the hypotheses to be sorted and we also want `beam.front()` to be the single-best in the end.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.hh
```diff
@@ -0,0 +1,153 @@
...
+        std::string toString() const;
+    };
+
+public:
+    static const Core::ParameterInt   paramMaxBeamSize;
+    static const Core::ParameterFloat paramScoreThreshold;
+    static const Core::ParameterInt   paramBlankLabelIndex;
+    static const Core::ParameterBool  paramAllowLabelLoop;
+    static const Core::ParameterBool  paramUseSentenceEnd;
+    static const Core::ParameterBool  paramSentenceEndIndex;
```
larissakl, line 81: These two are currently not used. I don't know if you want keep them if at some point you introduce sentence-end handling or if you want to remove them for now.

src/Speech/Makefile
```diff
@@ -46,6 +46,7 @@ CHECK_O		= $(OBJDIR)/check.o \
 		  ../Mm/libSprintMm.$(a) \
 		  ../Mc/libSprintMc.$(a) \
 		  ../Search/libSprintSearch.$(a) \
```
curufinwe, line 48: Better: use `LIBS_SEARCH` here and also for all other Makefiles below.

src/Tools/Archiver/Makefile
```diff
@@ -12,6 +12,7 @@ TARGETS		= archiver$(exe)
 ARCHIVER_O	= $(OBJDIR)/Archiver.o \
                  ../../Speech/libSprintSpeech.$(a) \
                  ../../Search/libSprintSearch.$(a) \
+				 ../../Search/LexiconfreeTimesyncBeamSearch/libSprintLexiconfreeTimesyncBeamSearch.$(a) \
```
curufinwe, line 15: correct indentation please.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.hh
```diff
@@ -0,0 +1,153 @@
...
+    bool  useScorePruning_;
+    Score scoreThreshold_;
+
+    bool           useBlank_;
+    Nn::LabelIndex blankLabelIndex_;
+
+    bool allowLabelLoop_;
+
+    bool logStepwiseStatistics_;
+    bool debugLogging_;
```
curufinwe, line 139: The usual pattern for this in RASR is to have a separate Channel object and if that is set log into it. For example here: https://github.com/rwth-i6/rasr/blob/639e9fafe767ad1ce517a86840088b280756d4d2/src/Core/Application.cc#L116 . Except you would not throw an error if the Channel is not open.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,447 @@
...
+ */
+
+#include "LexiconfreeTimesyncBeamSearch.hh"
+#include <Core/XmlStream.hh>
+#include <Lattice/LatticeAdaptor.hh>
+#include <Nn/LabelScorer/LabelScorer.hh>
+#include <Nn/LabelScorer/ScoringContext.hh>
+#include <algorithm>
+#include <strings.h>
+#include "Search/Traceback.hh"
```
curufinwe, line 23:
```suggestion
#include "LexiconfreeTimesyncBeamSearch.hh"

#include <algorithm>
#include <strings.h>

#include <Core/XmlStream.hh>
#include <Lattice/LatticeAdaptor.hh>
#include <Nn/LabelScorer/LabelScorer.hh>
#include <Nn/LabelScorer/ScoringContext.hh>
#include <Search/Traceback.hh>
```

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,447 @@
...
+        "Prune any hypotheses with a score that is at least this much worse than the best hypothesis. If not set, no score pruning will be done.",
+        Core::Type<Score>::max, 0);
+
+const Core::ParameterInt LexiconfreeTimesyncBeamSearch::paramBlankLabelIndex(
+        "blank-label-index",
+        "Index of the blank label in the lexicon. Can also be inferred from lexicon if it has a lemma with `special='blank'`. If not set, the search will not use blank.",
+        Core::Type<int>::max);
+
+const Core::ParameterBool LexiconfreeTimesyncBeamSearch::paramAllowLabelLoop(
+        "allow-label-loop",
```
curufinwe, line 49: Based on the description I would rather name this "collapse-repeated-labels".

curufinwe (changes requested):
Global comment: It would be nice if the order of method definitions would match the order of declarations in the class.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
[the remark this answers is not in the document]
larissakl: The order is needed in the recombination because it is assumed that the best hypothesis is first and if there is a hypothesis with the same scoring context later, we know that its score is worse.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,447 @@
...
+                     inferTransitionType(hyp.currentToken, tokenIdx),
+                     hypIndex});
+        }
+    }
+
+    /*
+     * Create scoring requests for the label scorer.
+     * Each extension candidate makes up a request.
+     */
+    std::vector<Nn::LabelScorer::Request> requests;
```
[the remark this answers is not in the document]
larissakl: We could maybe combine filling the requests and the extensions vector in the same loop? But I don't think we can completely avoid the extensions vector.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
larissakl: I see your point, but as we have a parameter "allow-label-loop" in the tree builder now it might be a good idea to have the same name here, because it basically controls the same thing.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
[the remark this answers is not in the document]
SimBe195: I can rewrite the pruning and recombination functions to not depend on sorted hypotheses.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,489 @@
...
+          recombinedHypotheses_(),
+          initializationTime_(),
+          featureProcessingTime_(),
+          scoringTime_(),
+          contextExtensionTime_(),
+          numHypsAfterScorePruning_("num-hyps-after-score-pruning"),
+          numHypsAfterBeamPruning_("num-hyps-after-beam-pruning"),
+          numActiveHyps_("num-active-hyps") {
+    beam_.reserve(maxBeamSize_);
+    extensions_.reserve(maxBeamSize_ * lexicon_->nLemmas());
```
larissakl, line 141: You can't use `lexicon_` here, it's not initialized yet but only set later in `setModelCombination`. At this point, `lexicon_->nLemmas()` equals 4294967294, so the max value of u32.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,489 @@
...
+          initializationTime_(),
+          featureProcessingTime_(),
+          scoringTime_(),
+          contextExtensionTime_(),
+          numHypsAfterScorePruning_("num-hyps-after-score-pruning"),
+          numHypsAfterBeamPruning_("num-hyps-after-beam-pruning"),
+          numActiveHyps_("num-active-hyps") {
+    beam_.reserve(maxBeamSize_);
+    newBeam_.reserve(maxBeamSize_);
+    requests_.reserve(extensions_.size());
```
larissakl, line 142: The reservation for `requests_` now has to be moved to `setModelCombination()` as well

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,499 @@
...
+
+LexiconfreeTimesyncBeamSearch::LexiconfreeTimesyncBeamSearch(Core::Configuration const& config)
+        : Core::Component(config),
+          SearchAlgorithmV2(config),
+          maxBeamSize_(paramMaxBeamSize(config)),
+          scoreThreshold_(paramScoreThreshold(config)),
+          blankLabelIndex_(paramBlankLabelIndex(config)),
+          collapseRepeatedLabels_(paramCollapseRepeatedLabels(config)),
+          logStepwiseStatistics_(paramLogStepwiseStatistics(config)),
+          debugChannel_(config, "debug", Core::Channel::standard),
```
larissakl, line 126:
```suggestion
          debugChannel_(config, "debug"),
```
Otherwise one would have to explicitly suppress the debug output.

larissakl (approved):

src/Search/Module.cc
```diff
@@ -101,6 +109,19 @@ SearchAlgorithm* Module_::createRecognizer(SearchType type, const Core::Configur
     return recognizer;
 }
 
+SearchAlgorithmV2* Module_::createSearchAlgorithmV2(const Core::Configuration& config) const {
+    SearchAlgorithmV2* searchAlgorithm = 0;
```
curufinwe, line 113:
```suggestion
    SearchAlgorithmV2* searchAlgorithm = nullptr;
```

src/Search/Traceback.cc
```diff
@@ -17,6 +17,9 @@
 
 #include <stack>
 
+#include <Lattice/LatticeAdaptor.hh>
+#include <Speech/Types.hh>
```
curufinwe, line 21: Are these really needed? You did not change anything in Traceback.cc so it should still compile fine without adding extra headers?

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.hh
```diff
@@ -0,0 +1,169 @@
...
+    Core::StopWatch initializationTime_;
+    Core::StopWatch featureProcessingTime_;
+    Core::StopWatch scoringTime_;
+    Core::StopWatch contextExtensionTime_;
+
+    Core::Statistics<u32> numHypsAfterScorePruning_;
+    Core::Statistics<u32> numHypsAfterBeamPruning_;
+    Core::Statistics<u32> numActiveHyps_;
+
+    bool finishedSegment_;
```
curufinwe, line 164: members before methods

curufinwe (approved):

PR #102, opened 2025-03-01
Modify AdvancedTreeSearch for CtcTreeBuilder support

larissakl:
Just minor adjustments to make the `AdvancedTreeSearch` work with the `CtcTreeBuilder`:

- Modified two if-statements related to the `stateDepths` which caused segmentation faults due to the self-loops in the CTC tree.

- Two assertions didn’t apply to the CTC tree, as its root state logic is significantly simpler. These assertions are now only checked if the `CtcTreeBuilder` is not used, so the previous behavior remains unchanged.


larissakl:
The `AdvancedTreeSearch` can be used with the `CtcTreeBuilder` and `RnaTreeBuilder` in two ways:
1. Enable loops (default setting) and set the TDP for loop to infinity
2. Disable loops (set `allow-label-loop` and `allow-blank-loop` to no) and set the TDP for loop to some value, e.g. 0.0

For both configurations, the TDP for skip has to be set to infinity.
Until now, the LM lookahead can only be used with option 2, with option 1 it will result in a segmentation fault.
Additionally, when label loops are disabled, the tree includes skip-transitions between identical labels, which are not allowed in the CTC-topology. This issue must be handled in the search code. I added a TODO in `AdvancedTreeSearch` for this.

larissakl:
Maybe you have a better idea for naming the config parameter "force-blank-between-identical-labels"...

src/Search/TreeBuilder.cc
```diff
@@ -1309,7 +1321,15 @@ StateId CtcTreeBuilder::extendState(StateId predecessor, StateTree::StateDesc de
 }
 
 void CtcTreeBuilder::addTransition(StateId predecessor, StateId successor) {
-    auto const& successorStateDesc = network_.structure.state(successor).stateDesc;
+    auto const& predecessorStateDesc = network_.structure.state(predecessor).stateDesc;
+    auto const& successorStateDesc   = network_.structure.state(successor).stateDesc;
+
+    if (forceBlank_) {
```
curufinwe, line 1327: Please don't add the check in a general helper function. If the function is called addTransition it should add a transition. It's better to add the check below, where we explicitly add the skip transition over blanks. Line 1366 in the original line numbering.

src/Search/TreeBuilder.cc
```diff
@@ -1422,7 +1448,13 @@ const Core::ParameterBool RnaTreeBuilder::paramLabelLoop(
         "allow label loops in the search tree",
         false);
 
+const Core::ParameterBool RnaTreeBuilder::paramForceBlank(
+        "force-blank-between-identical-labels",
```
curufinwe, line 1452: I would suggest: identical-labels -> repeated-labels

curufinwe (changes requested):

curufinwe (approved):

SimBe195 (approved):

curufinwe:
@Stefanwuu Once you can confirm that these changes work for you too we can merge. I'll set my review to request changes for this purpose only for now.

curufinwe (changes requested):
Waiting for confirmation that these changes are sufficient to run CTC decoding.

curufinwe (approved):

PR #103, opened 2025-03-04
StopWatch

SimBe195:
Add a simple `StopWatch` class to `Core`. This is a timer class with start/stop functions that accumulates all the timed intervals to a total.

The intended use is creating performance statistics for different subtasks in search. This PR already changes the `PerformanceCounter` in `AdvancedTreeSearch` to use this `StopWatch`. Other searches (e.g. #101) are also going to make use of it.


larissakl (approved):

src/Core/StopWatch.hh
```diff
@@ -0,0 +1,63 @@
...
+     */
+    double elapsedSeconds() const;
+    double elapsedCentiseconds() const;
+    double elapsedMilliseconds() const;
+    double elapsedMicroseconds() const;
+    double elapsedNanoseconds() const;
+
+private:
+    bool    running_;
+    timeval startTime_;
```
curufinwe, line 57: Let's use [https://en.cppreference.com/w/cpp/chrono/steady_clock](https://en.cppreference.com/w/cpp/chrono/steady_clock) instead of `timeval`.

src/Core/StopWatch.hh
```diff
@@ -0,0 +1,63 @@
...
+ * to a total.
+ */
+struct StopWatch {
+public:
+    StopWatch();
+
+    /*
+     * Stops timer if it is running and resets accumulated time to zero.
+     */
+    void reset();
```
curufinwe, line 33: Logically I would put reset after stop, but if you prefer this order I could also live with that.

curufinwe (changes requested):

src/Core/StopWatch.hh
SimBe195: The `TIMER_STOP` and `TIMER_START` macros work with `gettimeofday` and thus need `timeval` arguments. I used these macros to replicate the previous behavior of the `PerformanceCounter` in `AdvancedTreeSearch`. But I can use `steady_clock` instead of course; that's how I implemented it originally anyway before commit [5f82460](https://github.com/rwth-i6/rasr/pull/103/commits/5f82460b2ed00c0869aab6911da423b4ea810ae5).

curufinwe: Let's stick to steady_clock. I don't know how high the resolution is for gettimeofday, but it can change if the timezone is changed or somebody runs ntpdate on the machine. So we definetly should use a monotonic clock for benchmarking.

curufinwe (approved):

curufinwe:
Hmm, I guess I was a bit too slow renaming the PR.

PR #104, opened 2025-03-04
Add Search::LatticeTrace for traceback and lattice construction

SimBe195:
This PR adds a class `LatticeTrace` which is a `Tracebackitem` together with `predecessor` and `sibling` references in order to store the network of trace nodes making up a search lattice. This class is equipped with functions `performTraceback` and `buildWordLattice` which trace backwards from the node itself to build up a `Traceback` vector or `StandardWordLattice`. This class is going to be used in future search algorithms (e.g. #101).

Note: The [Trace class](https://github.com/rwth-i6/rasr/blob/21b9645a91d70e4c440e2924377d024dbb9ec90c/src/Search/AdvancedTreeSearch/Trace.hh#L52) of the `AdvancedTreeSearch` module is very similar but has a few additional members specific to that module. Because of the self-referential nature of the class (through `predecessor` and `sibling`) it is not straightforward to have the `AdvancedTreeSearch` `Trace` inherit from `LatticeTrace` though without introducing a bunch of `dynamic_cast` calls. So the options are:

- Keep `Trace` and `LatticeTrace` separate and accept that they are very similar
- Add the missing members from `Trace` to `LatticeTrace` to join them into one single class even though the other members are generally unused
- Make `Trace` inherit from `LatticeTrace` and use `dynamic_cast` wherever siblings or predecessors are accessed


src/Search/Traceback.hh
```diff
@@ -60,14 +67,97 @@ public:
...
+
+    /*
+     * Perform best-predecessor traceback.
+     * Ordered by increasing timestep.
+     */
+    Core::Ref<Traceback> performTraceback() const;
+
+    /*
+     * Build a word lattice from a traces. The given trace will be represent the final lattice
+     * state and it is traced back along predecessors and siblings until ending up at a trace with empty predecessor
```
larissakl, line 136:
```suggestion
     * Build a word lattice from the traces. The given trace will represent the final lattice
     * state and it is traced back along predecessors and siblings until ending up at a trace with empty predecessor
```

src/Search/Traceback.cc
```diff
@@ -74,4 +76,139 @@ Lattice::WordLatticeRef Traceback::wordLattice(Core::Ref<const Bliss::Lexicon> l
     result->setWordBoundaries(Core::ref(new Lattice::WordBoundaries));
     return result;
 }
+
+LatticeTrace::LatticeTrace(
+        Core::Ref<LatticeTrace> const&        predecessor,
+        const Bliss::LemmaPronunciation*      pronunciation,
+        Speech::TimeframeIndex                timeframe,
+        ScoreVector                           scores,
+        Search::TracebackItem::Transit const& transit)
```
larissakl, line 85:
```suggestion
        Transit const& transit)
```

larissakl (approved):

curufinwe (approved):

src/Search/Traceback.cc
```diff
@@ -74,4 +76,139 @@ Lattice::WordLatticeRef Traceback::wordLattice(Core::Ref<const Bliss::Lexicon> l
     result->setWordBoundaries(Core::ref(new Lattice::WordBoundaries));
     return result;
 }
+
+LatticeTrace::LatticeTrace(
+        Core::Ref<LatticeTrace> const&        predecessor,
+        const Bliss::LemmaPronunciation*      pronunciation,
+        Speech::TimeframeIndex                timeframe,
+        ScoreVector                           scores,
+        Transit const& transit)
```
larissakl, line 85: Sorry, that was my fault, the indentation is incorrect now

PR #105, opened 2025-03-05
Add VocabTextLexiconParser for simple text-based lexica

larissakl:
Support for simple text lexica containing the vocabulary/list of lemmas.
This is can for example be used in the Lexiconfree beam search (#101) for which the XML-format is unnecessarily complicated.
Instead of an XML-lexicon, one can use a txt-file containing one label per line.
For example

```
A
B
C
```

will be parsed to a Bliss Lexicon equivalent to

```
<?xml version="1.0" ?>
<lexicon>
  <phoneme-inventory>
    <phoneme>
      <symbol>A</symbol>
      <variation>none</variation>
    </phoneme>
    <phoneme>
      <symbol>B</symbol>
      <variation>none</variation>
    </phoneme>
    <phoneme>
      <symbol>C</symbol>
      <variation>none</variation>
    </phoneme>
    <phoneme>
 </phoneme-inventory>
  <lemma>
    <orth>A</orth>
    <phon>A</phon>
  </lemma>
  <lemma>
    <orth>B</orth>
    <phon>B</phon>
  </lemma>
  <lemma>
    <orth>C</orth>
    <phon>C</phon>
  </lemma>
</lexicon>
```

### DIFF
--- a/src/Bliss/Lexicon.cc
+++ b/src/Bliss/Lexicon.cc
@@ -136,16 +136,37 @@ Lexicon::~Lexicon() {
     delete internal_;
 }
 
+const Core::Choice Lexicon::lexiconTypeChoice(
+        "vocab-txt", vocabTxtLexicon,
+        "vocab-text", vocabTxtLexicon,
+        "xml", xmlLexicon,
+        Core::Choice::endMark());
+
+const Core::ParameterChoice Lexicon::paramLexiconType(
+        "type", &Lexicon::lexiconTypeChoice, "type of the lexicon file", xmlLexicon);
+
 void Lexicon::load(const std::string& filename) {
     Core::MD5 md5;
-    if (md5.updateFromFile(filename))
+    if (md5.updateFromFile(filename)) {
         dependency_.setValue(md5);
-    else
+    } else {
         warning("could not derive md5 sum from file '%s'", filename.c_str());
-    LexiconParser parser(config, this);
+    }
+    std::unique_ptr<LexiconParser> parser;
+    switch (paramLexiconType(config)) {
+    // text-based lexicon
+    case LexiconType::vocabTxtLexicon:
+        parser = std::make_unique<VocabTextLexiconParser>(this);
+        break;
+    // xml-based lexicon
+    case LexiconType::xmlLexicon:
+        parser = std::make_unique<XmlLexiconParser>(config, this);
+        break;
+    }
     log("reading lexicon from file") << " \"" << filename << "\" ...";
-    if (parser.parseFile(filename.c_str()) != 0)
+    if (!parser->parseFile(filename)) {
         error("Error while reading lexicon file.");
+    }
     log("dependency value: ") << dependency_.value();
 }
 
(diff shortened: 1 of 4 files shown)


src/Bliss/LexiconParser.cc
```diff
@@ -379,12 +379,90 @@ void LexiconParser::loadWhitelist(const Core::Configuration& config, Core::Strin
...
+
+// helper function to handle one label and create a corresponding phoneme
+void VocabTextLexiconParser::createPhoneme(const std::string& line) {
+    std::string symbol(line);
+    stripWhitespace(symbol);  // in case there are any unintentional whitespaces
+    suppressTrailingBlank(symbol);
+
+    // check if phoneme was already added (if one label appears more than once)
+    if (phonemeInventory_->phoneme(symbol)) {
+        return;
```
SimBe195, line 435: In my opinion rasr should not silently accept and skip duplicate symbols. This would shift all indices afterwards which is not expected by the user. I would rather throw an error.

src/Bliss/LexiconParser.cc
```diff
@@ -379,12 +379,90 @@ void LexiconParser::loadWhitelist(const Core::Configuration& config, Core::Strin
...
+void VocabTextLexiconParser::createLemmata() {
+    // iterate over the phonemes which were assigned to the inventory previously
+    auto phonemes = phonemeInventory_->phonemes();
+    for (auto it = phonemes.first; it != phonemes.second; ++it) {
+        const Phoneme* phoneme = *it;
+        std::string    symbol  = phoneme->symbol();
+
+        // check if lemma was already added (should not happen)
+        if (lexicon_->lemma(symbol)) {
+            return;
```
SimBe195, line 456: Better change this to a `verify` because if this happens something is wrong with the program.

SimBe195 (changes requested):

src/Bliss/Lexicon.cc
```diff
@@ -136,16 +136,38 @@ Lexicon::~Lexicon() {
     delete internal_;
 }
 
+const Core::Choice Lexicon::lexiconTypeChoice(
```
curufinwe, line 139: Before looking further into the details here one high-level suggestion: Different input formats can be handled via Format Specifiers. See the relevant section in https://github.com/rwth-i6/rasr/blob/master/doc/architecture.rst .

curufinwe (changes requested):

src/Bliss/LexiconParser.hh
```diff
@@ -116,12 +127,55 @@ private:
...
+
+/**
+ * Parser for text lexicon files containing the vocab, so only the labels
+ * This is meant for "lexicon-free" search
+ * The .txt-file should contain one label per line
+ */
+class VocabTextLexiconParser : public LexiconParser {
+private:
+    Lexicon*          lexicon_;
+    PhonemeInventory* phonemeInventory_;
```
SimBe195, line 156: These could be `Core::Ref` instead of raw pointers.

SimBe195 (changes requested):

curufinwe (changes requested):

curufinwe (approved):

SimBe195 (approved):

PR #106, opened 2025-03-05
Add Flf::RecognizerNodeV2

SimBe195:
Adds a new Flf node `RecognizerNodeV2`. This is a node that performs segment-wise recognition using a `SearchAlgorithmV2`. It is much more minimalistic than the original `RecognizerNode` and doesn't provide any additional functionality apart from running recognition and converting the result into an Flf lattice.

Depends on #101.


larissakl (approved):

src/Flf/RecognizerV2.cc
```diff
@@ -0,0 +1,242 @@
...
+
+void RecognizerNodeV2::init(std::vector<std::string> const& arguments) {
+    modelCombination_ = Core::ref(new Speech::ModelCombination(
+            config,
+            searchAlgorithm_->requiredModelCombination(),
+            searchAlgorithm_->requiredAcousticModel(),
+            Lexicon::us()));
+    searchAlgorithm_->setModelCombination(*modelCombination_);
+    if (not connected(0)) {
+        criticalError("Speech segment at port 1 required");
```
larissakl, line 215: Shouldn't it say port 0 here, not port 1?

SimBe195: Yes it should. Previously there was an optional lattice connection at port 0 which I removed. Fixed the message now.

src/Speech/ModelCombination.cc
```diff
@@ -92,6 +105,10 @@ void ModelCombination::setLanguageModel(Core::Ref<Lm::ScaledLanguageModel> langu
         languageModel_->setParentScale(scale());
 }
 
+void ModelCombination::setLabelScorer(Core::Ref<Nn::LabelScorer> ls) {
+    labelScorer_ = ls;
+}
+
```
curufinwe, line 111: This is already implemented in the header file.

src/Flf/RecognizerV2.hh
```diff
@@ -0,0 +1,74 @@
...
+    void work();
+
+    /*
+     * Convert an output lattice from `searchAlgorithm_` to an Flf lattice
+     */
+    ConstLatticeRef buildLattice(Core::Ref<const Search::LatticeAdaptor> latticeAdaptor, std::string segmentName);
+
+    std::pair<ConstLatticeRef, ConstSegmentRef> resultBuffer_;
+
+    Search::SearchAlgorithmV2*          searchAlgorithm_;
```
curufinwe, line 67: Why isn't this a `std::unique_ptr`?

src/Flf/RecognizerV2.cc
```diff
@@ -0,0 +1,241 @@
...
+            searchAlgorithm_->requiredAcousticModel(),
+            Lexicon::us()));
+    searchAlgorithm_->setModelCombination(*modelCombination_);
+    if (not connected(0)) {
+        criticalError("Speech segment at port 0 required");
+    }
+}
+
+void RecognizerNodeV2::sync() {
+    resultBuffer_.first.reset();
```
curufinwe, line 219: These are Const...Ref objects. How come you can call reset on them?

src/Flf/RecognizerV2.cc
```diff
@@ -0,0 +1,241 @@
...
+
+    // Initialize recognizer and feature extractor
+    searchAlgorithm_->reset();
+    searchAlgorithm_->enterSegment();
+
+    featureExtractor_->enterSegment(segment);
+    DataSourceRef dataSource = featureExtractor_->extractor();
+    dataSource->initialize(const_cast<Bliss::SpeechSegment*>(segment));
+    FeatureRef feature;
+    dataSource->getData(feature);
```
curufinwe, line 53: Getting the first feature should be part of the timed section. Otherwise if you use tf-fwd node (or something similar) the bulk of the work goes unreported.

curufinwe (changes requested):
Only minor stuff. Did not check lattice building code as I assume you copied that with only small modifications from AdvancedTreeSearch, correct?

src/Flf/RecognizerV2.cc
SimBe195: A void Ref means that the buffer is not filled and `work()` must be called (see `sendSegment` and `sendLattice` functions). The call of `sync()` clears it again by resetting the Refs back to void. This behavior is copied from the original `RecognizerNode`.

curufinwe: Nvm, I thought that this is a method of the Lattice/Segment object that you call, but it's a method of Core::Ref. So everything is fine here.

curufinwe (approved):

PR #108, opened 2025-03-17
Fix CTCTopologyGraphBuilder: force blank label between two consecutive identical labels

Stefanwuu:
When training CTC model with FSA created by AllophoneStateGraphBuilder, we observe around 5% relative WER degradation compared to training with built-in Pytorch CTC loss. We found out this is due to incorrect handling of repeated identical labels(e.g. aa in aachen when using character label unit). Without a forced blank in between, CTC topology can't differentiate loop and forward between repeated identical labels. After using the fixed version, we are able to get the same WER.

### DIFF
--- a/src/Speech/AllophoneStateGraphBuilder.cc
+++ b/src/Speech/AllophoneStateGraphBuilder.cc
@@ -467,7 +467,7 @@ Fsa::ConstAutomatonRef CTCTopologyGraphBuilder::buildTransducer(Fsa::ConstAutoma
     Fsa::ConstAutomatonRef model = buildFlatTransducer(lemmaAcceptor);
     model = addLoopTransition(model);
     // remove epsilon so that repeated identical label detection could work
-    model = Fsa::removeEpsilons(Fsa::removeDisambiguationSymbols(Fsa::projectInput(model)));
+    model = Fsa::removeEpsilons(Fsa::removeDisambiguationSymbols(model));
     Core::Ref<Fsa::StaticAutomaton> automaton = Fsa::staticCopy(model);
 
     finalStateId_ = Core::Type<Fsa::StateId>::max;
--- a/src/Speech/AllophoneStateGraphBuilder.cc
+++ b/src/Speech/AllophoneStateGraphBuilder.cc
@@ -500,9 +500,9 @@ void CTCTopologyGraphBuilder::addBlank(Core::Ref<Fsa::StaticAutomaton>& automato
     // find non-blank loop label for later consecutive identical label handling
     Fsa::LabelId loopLabel = Fsa::InvalidLabelId;
     for (u32 idx = 0; idx < nArcs; ++idx) {
-        const Fsa::Arc* arcLoop = state->getArc(idx);
-        if ((arcLoop->target_ == s) && (arcLoop->input_ != blankId_)) {
-            loopLabel = arcLoop->input_;
+        const Fsa::Arc* arc = state->getArc(idx);
+        if (arc->target_ == s && arc->input_ != blankId_) {
+            loopLabel = arc->input_;
             break;
         }
     }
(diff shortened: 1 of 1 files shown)


src/Speech/AllophoneStateGraphBuilder.cc
```diff
@@ -495,6 +497,15 @@ void CTCTopologyGraphBuilder::addBlank(Core::Ref<Fsa::StaticAutomaton>& automato
     Fsa::Weight zeroWeight(0);
     Fsa::State* state = automaton->fastState(s);
     u32 nArcs = state->nArcs();
+    // find non-blank loop label for later consecutive identical label handling
+    Fsa::LabelId loopLabel = Fsa::InvalidLabelId;
+    for (u32 idx = 0; idx < nArcs; ++idx) {
+        const Fsa::Arc* arcLoop = state->getArc(idx);
+        if ((arcLoop->target_ == s) && (arcLoop->input_ != blankId_)) {
+            loopLabel = arcLoop->input_;
+            break;
```
albertz, line 506: I wonder a bit: Is this really always unique?
Instead of the `break`, you could maybe also continue the loop and assert that you don't get another label?

src/Speech/AllophoneStateGraphBuilder.cc
```diff
@@ -495,6 +497,15 @@ void CTCTopologyGraphBuilder::addBlank(Core::Ref<Fsa::StaticAutomaton>& automato
     Fsa::Weight zeroWeight(0);
     Fsa::State* state = automaton->fastState(s);
     u32 nArcs = state->nArcs();
+    // find non-blank loop label for later consecutive identical label handling
+    Fsa::LabelId loopLabel = Fsa::InvalidLabelId;
+    for (u32 idx = 0; idx < nArcs; ++idx) {
+        const Fsa::Arc* arcLoop = state->getArc(idx);
```
curufinwe, line 503: `arcLoop` -> `arc` (not every arc considered here is a loop, thus I find the variable name a bit confusing)

src/Speech/AllophoneStateGraphBuilder.cc
```diff
@@ -509,7 +520,18 @@ void CTCTopologyGraphBuilder::addBlank(Core::Ref<Fsa::StaticAutomaton>& automato
         Fsa::StateId blankStateId = automaton->maxStateId();
         blankState->newArc(blankStateId, zeroWeight, blankId_, Fsa::Epsilon);
         blankState->newArc(target, a->weight_, input, a->output_);
-        state->newArc(blankStateId, zeroWeight, blankId_, Fsa::Epsilon); // invalidate a
+        // handle consecutive identical label: if label loop and forward represent the same label,
+        // we should overwrite the original arc target to make the blank unskippable
+        if (loopLabel != Fsa::InvalidLabelId && 
+            acousticModel_->emissionIndex(input) == acousticModel_->emissionIndex(loopLabel)) {
+            Fsa::Arc* aa = const_cast<Fsa::Arc*>(state->getArc(idx));
```
curufinwe, line 527: How about renaming `aa` to `arc`?

src/Speech/AllophoneStateGraphBuilder.cc
```diff
@@ -466,6 +466,8 @@ Fsa::ConstAutomatonRef CTCTopologyGraphBuilder::addLoopTransition(Fsa::ConstAuto
 Fsa::ConstAutomatonRef CTCTopologyGraphBuilder::buildTransducer(Fsa::ConstAutomatonRef lemmaAcceptor) {
     Fsa::ConstAutomatonRef model = buildFlatTransducer(lemmaAcceptor);
     model = addLoopTransition(model);
+    // remove epsilon so that repeated identical label detection could work
+    model = Fsa::removeEpsilons(Fsa::removeDisambiguationSymbols(Fsa::projectInput(model)));
```
curufinwe, line 470: Why is the `projectInput` operation required here?

curufinwe (changes requested):

albertz (approved):

curufinwe (approved):

PR #110, opened 2025-03-25
Remove a lot of unused neural network code

curufinwe:
I was in the mood of deleting code and started with some of the Neural-Network code in RASR that I think nobody has used in many many years.

Points of discussion:
 - @albertz I removed CTC criterion computation from RASR, I guess you are not using it since many years, but I want to ask anyway
 - @michelwi Do you think the changes I made have an impact on seq-discriminative training? I think not, but maybe you could double-check, or give me a small setup to verify?
 - @SimBe195 @Marvin84 : Do you see anything here that you might still be using?


michelwi (approved):
> Do you think the changes I made have an impact on seq-discriminative training?

I don't think sequence discriminative training will be affected.

SimBe195 (approved):
I also don't use any of this.

albertz (approved):
Looks good to me.

PR #112, opened 2025-03-27
Add DataView class as shared_ptr adapter for LabelScorer inputs

SimBe195:
Rewrite input logic of `LabelScorer` and input/output logic of `Encoder` by introducing a class `DataView` instead of using `std::shared_ptr` directly.

This `DataView` class encapsulates a `std::shared_ptr` and a size. The internal shared_ptr is tied to the lifetime of the original data container in order to make sure it stays valid as long as the `DataView` object is alive. It can be initialized using various data containers such as a `Core::Ref<Mm::Feature::Vector>`, a `shared_ptr`, an `Onnx::Value` or a `pybind11::array_t`.

This way the lifetime management of the original data containers and `shared_ptr` conversion are all collected into one place.


src/Nn/LabelScorer/Makefile
```diff
@@ -12,6 +12,7 @@ TARGETS		= libSprintLabelScorer.$(a)
 LIBSPRINTLABELSCORER_O =  \
     $(OBJDIR)/BufferedLabelScorer.o \
     $(OBJDIR)/CombineLabelScorer.o \
+	$(OBJDIR)/DataView.o \
```
larissakl, line 15: Indentation

larissakl (approved):

curufinwe:
I cleaned up some build related stuff. Tested it in this updated apptainer: https://github.com/rwth-i6/rasr/pull/122 . But that's mostly for up-to-date versions of TF & ONNX + support for FFMPEG and a working pybind11 setup, so it should also work for you @SimBe195 . I'll merge this for now to make progress with the other PRs. If there is issues with your build let's do another PR.

PR #114, opened 2025-04-09
Add cache-cleanup functionality to LabelScorer

SimBe195:
This introduces a function `cleanupCaches` to the `LabelScorer` class which - given a collection of active `ScoringContext` objects - is supposed to clean up any internal buffers or caches by removing values which are no longer needed (e.g. corresponding to outdated timeframes).

This is especially important for long-form audio or streaming so that the caches don't grow indefinitely.

In the `LexiconfreeTimesyncBeamSearch` it is called after every decode step with the scoring contexts of all active hypotheses in the beam.

Depends on #112.


src/Nn/LabelScorer/CombineLabelScorer.cc
```diff
@@ -71,6 +71,12 @@ ScoringContextRef CombineLabelScorer::extendedScoringContext(Request const& requ
     return Core::ref(new CombineScoringContext(std::move(extScoringContexts)));
 }
 
+void CombineLabelScorer::cleanupCaches(Core::CollapsedVector<ScoringContextRef> const& activeContexts) {
+    for (auto& scaledScorer : scaledScorers_) {
+        scaledScorer.scorer->cleanupCaches(activeContexts);
```
larissakl, line 76: Maybe I'm wrong, but don't you need to cast `activeContexts` to `CombineScoringContext` and iterate over the scoring contexts here?

SimBe195: No that's absolutely right. The `activeContexts` can't be directly fed to the sub scorer. First, the appropriate sub-ScoringContext needs to be extracted. Fixed now.

src/Nn/LabelScorer/BufferedLabelScorer.cc
```diff
@@ -37,4 +39,30 @@ void BufferedLabelScorer::addInput(DataView const& input) {
...
+
+    auto minActiveTime = minActiveTimeIndex(activeContexts);
+    if (minActiveTime > numDeletedInputs_) {
+        size_t deleteInputs = minActiveTime - numDeletedInputs_;
+        deleteInputs        = std::min(deleteInputs, inputBuffer_.size());
+        inputBuffer_.erase(inputBuffer_.begin(), inputBuffer_.begin() + deleteInputs);
+        numDeletedInputs_ += deleteInputs;
+    }
+}
+std::optional<DataView> BufferedLabelScorer ::getInput(Speech::TimeframeIndex timeIndex) const {
```
larissakl, line 55:
```suggestion
std::optional<DataView> BufferedLabelScorer::getInput(Speech::TimeframeIndex timeIndex) const {
```

src/Nn/LabelScorer/BufferedLabelScorer.hh
```diff
@@ -42,9 +44,18 @@ public:
...
-    std::vector<DataView> inputBuffer_;         // Buffer that contains all the feature data for the current segment
-    bool                  expectMoreFeatures_;  // Flag to record segment end signal
+    bool expectMoreFeatures_;  // Flag to record segment end signal
+
+    virtual Speech::TimeframeIndex minActiveTimeIndex(Core::CollapsedVector<ScoringContextRef> const& activeContexts) const = 0;
+    std::optional<DataView>        getInput(Speech::TimeframeIndex timeIndex) const;
+
+private:
+    std::deque<DataView> inputBuffer_;       // Buffer that contains all the feature data for the current segment
+    size_t               numDeletedInputs_;  // Count deleted inputs in order to adress the correct index in inputBuffer_
```
larissakl, line 58:
```suggestion
    size_t               numDeletedInputs_;  // Count deleted inputs in order to address the correct index in inputBuffer_
```

larissakl (approved):

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -337,22 +338,31 @@ bool LexiconfreeTimesyncBeamSearch::decodeStep() {
     recombination(newBeam_);
     numActiveHyps_ += newBeam_.size();
 
-    if (logStepwiseStatistics_) {
-        clog() << Core::XmlFull("active-hyps", newBeam_.size());
+    /*
+     * Clean up label scorer caches.
+     */
+    Core::CollapsedVector<Nn::ScoringContextRef> activeContexts;
```
curufinwe, line 344: Consider doing this only every N frames.

curufinwe (changes requested):

curufinwe (approved):

PR #115, opened 2025-04-09
Add Nn::NoContextOnnxLabelScorer

SimBe195:
This adds a new `LabelScorer` class `NoCtxOnnxLabelScorer`. This scorer works by forwarding only the input feature at the current timestep through an ONNX model, without any label history.
This is suitable for example for CTC outputs consisting of a linear layer + -log_softmax activation.

If the CTC output is the only output, the encoder and output layer can be put together into an "encoder-only" label scorer. However, when the CTC output is one of several outputs based on a shared encoder, the CTC output head must be separated from the encoder. The `NoCtxOnnxLabelScorer` can be used for this purpose.

Depends on #114.


larissakl (approved):

src/Core/Hash.hh
```diff
@@ -15,6 +15,7 @@
 #ifndef _CORE_HASH_HH
 #define _CORE_HASH_HH
 
+#include <cstdint>
```
curufinwe, line 18: This should not be needed

src/Core/CollapsedVector.hh
```diff
@@ -124,6 +130,26 @@ inline const T& CollapsedVector<T>::front() const {
     return data_.front();
 }
 
+template<typename T>
```
curufinwe, line 133: Didn't we want to not have iterators into the physical vector? If you have iterators they should respect logical size.

src/Nn/LabelScorer/NoCtxOnnxLabelScorer.hh
```diff
@@ -0,0 +1,77 @@
...
+/*
+ * A LabelScorer that computes scores by forwarding only the input feature at the current timestep through an ONNX model,
+ * without any label history.
+ * This is suitable for example for CTC outputs consisting of a linear layer + -log_softmax activation.
+ *
+ * If the CTC output is the only output, the encoder and output layer can be put together into an "encoder-only" label scorer.
+ * However, when the CTC output is one of several outputs based on a shared encoder, the CTC output head must be separated
+ * from the encoder. The NoCtxOnnxLabelScorer can be used for this purpose.
+ */
+class NoCtxOnnxLabelScorer : public BufferedLabelScorer {
```
curufinwe, line 34:
```suggestion
class NoContextOnnxLabelScorer : public BufferedLabelScorer {
```

src/Nn/LabelScorer/BufferedLabelScorer.hh
```diff
@@ -16,8 +16,10 @@
 #ifndef BUFFERED_LABEL_SCORER_HH
 #define BUFFERED_LABEL_SCORER_HH
 
+#include <deque>
 #include "DataView.hh"
```
curufinwe, line 20:
```suggestion
#include <deque>

#include "DataView.hh"
```

curufinwe (changes requested):
Only minor changes required.

PR #116, opened 2025-04-09
Add Nn::FixedContextOnnxLabelScorer

SimBe195:
This adds a new `LabelScorer` class `LimitedCtxOnnxLabelScorer`. This scorer works by forwarding the input feature at the current timestep together with a fixed-size sequence of history tokens through an ONNX model.

A common use case would be a neural transducer model with a fixed-size history.

Depends on #115.


larissakl (approved):

curufinwe (changes requested):
Only minor issues

curufinwe (approved):

PR #117, opened 2025-04-10
Python bindings for Search Algorithm

SimBe195:
This PR adds python bindings in `LibRASR` that can run recognition using a `SearchAlgorithmV2`. These bindings can be used for example to perform RASR search in a regular `ReturnnForwardJob`.

Depends on #112.

### DIFF
--- a/src/Python/Search.cc
+++ b/src/Python/Search.cc
@@ -1,0 +1,106 @@
+/** Copyright 2025 RWTH Aachen University. All rights reserved.
+ *
+ *  Licensed under the RWTH ASR License (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.hltpr.rwth-aachen.de/rwth-asr/rwth-asr-license.html
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#include "Search.hh"
+
+#include <Search/Module.hh>
+#include <Speech/ModelCombination.hh>
+
+namespace py = pybind11;
+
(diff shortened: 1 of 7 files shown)


src/Python/Search.cc
```diff
@@ -0,0 +1,106 @@
...
+
+    auto                       traceback = searchAlgorithm_->getCurrentBestTraceback();
+    std::vector<TracebackItem> result;
+    result.reserve(traceback->size());
+
+    u32 prevTime = 0;
+
+    for (auto it = traceback->begin(); it != traceback->end(); ++it) {
+        result.push_back({
+                it->pronunciation->lemma()->symbol(),
```
larissakl, line 89:
```suggestion
    for (auto it = traceback->begin(); it != traceback->end(); ++it) {
        std::string lemma = "";
        if (it->pronunciation and it->pronunciation->lemma()) {
            lemma = std::string(it->pronunciation->lemma()->symbol().str());
        }
        result.push_back({
                lemma,
```

This caused a segmentation fault

larissakl (changes requested):
When I tested the new search algorithms with these Python bindings I ran into a few problems.

larissakl (approved):

src/Tools/LibRASR/Search.cc
```diff
@@ -0,0 +1,116 @@
...
+     * ========================
+     * === Search Algorithm ===
+     * ========================
+     */
+    py::class_<SearchAlgorithm> pySearchAlgorithm(
+            module,
+            "SearchAlgorithm",
+            "Class that can perform recognition using RASR.\n\n"
+            "The search algorithm is configured with a RASR config object.\n"
+            "It works by calling `enter_segment()`, passing segment features\n"
```
curufinwe, line 68: It would be helpful to also add here that the user has to call reset() before starting a new recognition.

src/Core/Hash.hh
```diff
@@ -15,6 +15,7 @@
 #ifndef _CORE_HASH_HH
 #define _CORE_HASH_HH
 
+#include <cstdint>
```
curufinwe, line 18: again this include!

curufinwe (changes requested):
I tested the code and it works well! Only some minor changes needed.

PR #118, opened 2025-04-11
Remove WordConditionedTreeSearch

larissakl:
As discussed, I have deleted the `WordConditionedTreeSearch` class, incl. the LanguageModelLookahead it used.

The class `RecognizerWithConstrainedLanguageModel` in `src/Speech/AdvancedLatticeExtractor` was still using it, I updated this to use the AdvancedTreeSearch instead. This class is used by `NumeratorLatticeGenerator` and `LatticeSetGenerator` (for acoustic rescoring). However, I'm not sure whether these components are still in active use and I don't have a setup to test if the AdvancedTreeSearch works correctly in this context, so please take a close look at that.


src/Lattice/Lattice.hh
```diff
@@ -262,9 +262,9 @@ public:
 /**
  * A "normal" word lattice with acoustic and language model scores
  * and word boundary information as it is produced by
- * WordConditionedTreeSearch.  This (badly named) class
+ * AdvancedTreeSearch.  This (badly named) class
  * facilitates the creation of word graph in
- * WordConditionedTreeSearch.  It will probably disappear in the
+ * AdvancedTreeSearch.  It will probably disappear in the
  * near future, when the WordLattice interface is refactored.
```
SimBe195, line 268: > It will probably disappear in the near future, when the WordLattice interface is refactored.

Hahaha

SimBe195 (approved):

src/Lattice/Lattice.hh
larissakl: I thought about deleting this sentence but I really liked this optimism :)

curufinwe:
@michelwi You know what RASR question I always ask you: could these changes impact seq. discriminative training (via the Speech/AdvancedLatticeExtractor.cc changes)?

src/Search/LanguageModelLookahead.cc
michelwi, line 1: I hope there is a different implementation of LM lookahead that is used in AdvancedTreeSearch 🙈

michelwi (approved):
> @michelwi You know what RASR question I always ask you: could these changes impact seq. discriminative training (via the Speech/AdvancedLatticeExtractor.cc changes)?

As discussed in some meeting: our sequence discriminative training pipeline uses the (non-flf) Recognizer with AdvancedTreeSearch (cf [here](https://github.com/rwth-i6/i6_core/blob/b3cc3482a5c5a546cf05433e39b826028dcf7f6e/discriminative_training/lattice_generation.py#L311)), which is still kept intact. So it should be fine 👍

src/Search/LanguageModelLookahead.cc
larissakl: Yes, in PR #119 I moved the LM lookahead from AdvancedTreeSearch to Search. The lookahead I deleted here was based on a StateTree and not on a HMMStateNetwork.

curufinwe (approved):

PR #121, opened 2025-04-23
Update Bliss::Lexicon & some related classes to return Core::Status

curufinwe:


src/Core/XmlParser.cc
```diff
@@ -342,22 +336,30 @@ int XmlParser::parse() {
...
 }
 
 int XmlParser::parseString(const char* str) {
     ctxt_ = xmlCreateMemoryParserCtxt(const_cast<char*>(str), strlen(str));
-    if (ctxt_ == NULL)
+    if (ctxt_ == NULL) {
         error("Failed to setup XML parser");
+        return -1;
+    }
     int status = parse();
```
SimBe195, line 363: `parse` and `parseString` could also be refactored to return `Core::Status` now but that's a separate thing.

SimBe195 (changes requested):

hannah220 (approved):

PR #123, opened 2025-05-05
Update windowing related functions to better support Pytorch compatible feature extraction

curufinwe:


src/Signal/Window.cc
```diff
@@ -95,26 +117,40 @@ const Core::ParameterBool WindowNode::paramFlushBeforeGap(
 
 WindowNode::WindowNode(const Core::Configuration& c)
         : Component(c), Predecessor(c) {
-    setWindowFunction(WindowFunction::create((WindowFunction::Type)WindowFunction::paramType(c)));
+    setWindowFunction(std::unique_ptr<WindowFunction>(WindowFunction::create((WindowFunction::Type)WindowFunction::paramType(c))));
```
SimBe195, line 120:
```suggestion
    setWindowFunction(std::unique_ptr<WindowFunction>(WindowFunction::create(static_cast<WindowFunction::Type>(WindowFunction::paramType(c)))));
```
C++ style cast instead of C style cast.

src/Signal/Window.cc
```diff
@@ -95,26 +117,40 @@ const Core::ParameterBool WindowNode::paramFlushBeforeGap(
...
     setFlushAll(paramFlushAll(c));
     setFlushBeforeGap(paramFlushBeforeGap(c));
 }
 
 bool WindowNode::setParameter(const std::string& name, const std::string& value) {
-    if (WindowFunction::paramType.match(name))
-        setWindowFunction(WindowFunction::create((WindowFunction::Type)WindowFunction::paramType(value)));
-    else if (paramShift.match(name))
+    if (WindowFunction::paramType.match(name)) {
+        setWindowFunction(std::unique_ptr<WindowFunction>(WindowFunction::create((WindowFunction::Type)WindowFunction::paramType(value))));
```
SimBe195, line 131:
```suggestion
        setWindowFunction(std::unique_ptr<WindowFunction>(WindowFunction::create(static_cast<WindowFunction::Type>WindowFunction::paramType(value)))));
```
Again C++ cast.

src/Signal/WindowFunction.hh
```diff
@@ -76,14 +77,16 @@ public:
...
 template<class Iterator>
 bool WindowFunction::work(const Iterator& begin, const Iterator& end) {
-    ensure(std::distance(begin, end) >= (s32)length());
-
-    if (needInit_ && !init())
+    if (needInit_ && !init()) {
         return false;
+    }
+
+    size_t effective_window = std::min<size_t>(std::distance(begin, end), length());
```
SimBe195, line 84:
```suggestion
    size_t effectiveWindow = std::min<size_t>(std::distance(begin, end), length());
```

src/Signal/WindowFunction.hh
```diff
@@ -76,14 +77,16 @@ public:
...
 
-    std::transform(window_.begin(), window_.end(), begin, begin, std::multiplies<Float>());
+    std::transform(window_.begin(), window_.begin() + effective_window, begin, begin, std::multiplies<Float>());
```
SimBe195, line 86:
```suggestion
    std::transform(window_.begin(), window_.begin() + effectiveWindow, begin, begin, std::multiplies<Float>());
```

src/Signal/WindowFunction.hh
```diff
@@ -76,14 +77,16 @@ public:
...
 
-    std::fill(begin + length(), end, 0);
+    // disregard samples that do not fit in window
+    std::fill(begin + effective_window, end, 0.0);
```
SimBe195, line 89:
```suggestion
    std::fill(begin + effectiveWindow, end, 0.0);
```

src/Signal/Window.cc
```diff
@@ -62,30 +67,47 @@ void Window::setSampleRate(f64 sampleRate) {
...
     setShift((u32)rint(shiftInS_ * sampleRate()));
 
     Predecessor::init();
 }
 
 void Window::transform(Vector<Sample>& out) {
-    windowFunction_->setLength(out.size());
+    u32 offset = (u32)rint(windowOffsetInS_ * sampleRate());
+    offset     = std::min<u32>(out.size(), offset);
+
```
hannah220, line 87: Should we ensure that `inputLengthInS_ > lengthInS_` ?
Otherwise windowFunction expects larger size of vector than it gets.

src/Signal/Window.cc
```diff
@@ -62,30 +67,47 @@ void Window::setSampleRate(f64 sampleRate) {
...
+    u32 windowLength = (u32)rint(lengthInS_ * sampleRate());
+    windowFunction_->setLength(windowLength);
 
-    if (!windowFunction_->work(out.begin(), out.end()))
+    std::fill(out.begin(), out.begin() + offset, 0.0);
```
hannah220, line 91: Also to make sure that `windowOffsetInS_ < inputLengthInS_` so that `out.begin() + offset` will not be larger than `out.end()`

src/Signal/WindowFunction.cc
```diff
@@ -102,12 +105,22 @@ bool HammingWindowFunction::init() {
...
+        for (u32 n = 0; n < window_.size(); n++) {
+            window_[n] = 0.5 - 0.5 * cos(2.0 * M_PI * n / M);
+        }
+    }
+    else {
+        u32 M = window_.size() - 1;
+        for (u32 n = 0; n <= M / 2; n++) {
+            window_[n] = window_[M - n] = 0.5 - 0.5 * cos(2.0 * M_PI * n / M);
+        }
+    }
```
SimBe195, line 123: Those two branches are nearly equivalent, just with a different value for `M`. We always have
$$\cos\left(\frac{2 \pi (M-n)}{M}\right) = \cos\left(2 \pi - \frac{2 \pi n}{M}\right) = \cos\left( - \frac{2 \pi n}{M}\right) = \cos\left(\frac{2 \pi n}{M}\right)$$
so the two "versions" of the for-loop achieve the same thing. So essentially, the for-loop can be moved out of the if, only the initialization of `M` needs to be inside. This has the effect that `0 == window_[0] == window_[length()-1]` for `periodic_ = false` and `0 == window_[0] != window_[length()-1]` for `periodic_ = true`. Is this how the logic of the window function is intended? Maybe some documentation explaining the difference between "hanning" and "periodic-hanning" would be good in this context.

SimBe195 (changes requested):

src/Signal/Window.cc
```diff
@@ -62,30 +67,47 @@ void Window::setSampleRate(f64 sampleRate) {
...
 // WindowNode
 /////////////
 
 const Core::ParameterFloat WindowNode::paramShift(
-        "shift", "shift of window");
+        "shift", "shift of window", 0.0, 0.0);
 
 const Core::ParameterFloat WindowNode::paramLength(
-        "length", "length of window");
+        "length", "length of window", 0.0, 0.0);
```
hannah220, line 104: Can we add something like, `window function is applied on this length`.

hannah220 (changes requested):

src/Signal/WindowFunction.cc
curufinwe: The reason I added this is because the default for Pytorch is periodic = true and that results in slightly different features: https://docs.pytorch.org/docs/stable/generated/torch.hann_window.html

src/Signal/Window.cc
curufinwe: PS: Actually this is taken care of in line 86.

hannah220 (approved):

PR #126, opened 2025-05-12
Add lexicon-free label-synchronous beam-search algorithm

SimBe195:
Implement a lexicon-free label-synchronous search algorithm (based on `SearchAlgorithmV2`). This is mostly similar to `LexiconfreeTimesyncBeamSearch` with the most notable differences being:
 - Instead of a blank token, this search needs a sentence-end token.
 - There is a distinction between terminated hypotheses and active hypotheses, depending on whether they have produced sentence-end yet.
 - Since hypotheses may be terminated at different lengths, length normalization is introduced for score comparison.
 - The search terminates when all hypotheses in the beam are terminated or when an upper limit of decoding steps has been reached. This limit depends on the number of input features that have been fed.


src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,594 @@
...
+            clog() << Core::XmlFull("num-hyps-after-score-pruning", beam_.size());
+        }
+    }
+
+    beamSizePruning();
+    numHypsAfterBeamPruning_ += beam_.size();
+    if (logStepwiseStatistics_) {
+        clog() << Core::XmlFull("num-hyps-after-beam-pruning", beam_.size());
+    }
+
```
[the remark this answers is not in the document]
SimBe195: Good catch, this is just an error. Fixed now.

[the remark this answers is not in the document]
larissakl: Ok good, but I think the comment "Prune terminated hypotheses among each other" does not fit then. `newBeam_` contains all terminated and active hypotheses.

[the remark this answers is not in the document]
SimBe195: Yes, I used to handle terminated hypotheses separately, this comment doesn't fit anymore. Fixed now.

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,594 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#include "LexiconfreeLabelsyncBeamSearch.hh"
+
+#include <algorithm>
+#include <strings.h>
```
larissakl, line 19:
```suggestion
#include <algorithm>
#include <numeric>
#include <strings.h>
```
Needed for std::accumulate

src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -84,6 +84,7 @@ public:
         BLANK_LOOP,
         INITIAL_LABEL,
         INITIAL_BLANK,
+        SENTENCE_END
```
larissakl, line 87: This also has to be included as a case in the switch-case of `LexiconfreeTimesyncBeamSearch` now

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,594 @@
...
+          scoringTime_(),
+          contextExtensionTime_(),
+          numHypsAfterScorePruning_("num-hyps-after-score-pruning"),
+          numHypsAfterBeamPruning_("num-hyps-after-beam-pruning"),
+          currentSearchStep_(0ul),
+          totalTimesteps_(0ul),
+          finishedSegment_(false) {
+    beam_.reserve(maxBeamSize_);
+    newBeam_.reserve(maxBeamSize_ * 2);  // terminated + active
+    recombinedHypotheses_.reserve(maxBeamSize_);
```
larissakl, line 142:
```suggestion
    recombinedHypotheses_.reserve(maxBeamSize_ * 2);
```
`recombination()` is called before the joint pruning and if no hypotheses are recombined, this vector has the same size as `newBeam_`

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,594 @@
...
+    numHypsAfterScorePruning_.write(clog());
+    numHypsAfterBeamPruning_.write(clog());
+}
+
+void LexiconfreeLabelsyncBeamSearch::beamSizePruningExtensions() {
+    if (extensions_.size() <= maxBeamSize_) {
+        return;
+    }
+
+    // Reorder the hypotheses by associated score value such that the first `beamSizeActive_` elements are the best
```
larissakl, line 508:
```suggestion
    // Reorder the hypotheses by associated score value such that the first `maxBeamSize_` elements are the best
```

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,594 @@
...
+    std::nth_element(extensions_.begin(), extensions_.begin() + maxBeamSize_, extensions_.end());
+    extensions_.resize(maxBeamSize_);  // Get rid of excessive elements
+}
+
+void LexiconfreeLabelsyncBeamSearch::beamSizePruning() {
+    if (newBeam_.size() <= maxBeamSize_) {
+        return;
+    }
+
+    // Reorder the hypotheses by associated score value such that the first `beamSizeTerminated_` elements are the best
```
larissakl, line 518:
```suggestion
    // Reorder the hypotheses by associated score value such that the first `maxBeamSize_` elements are the best
```

larissakl (changes requested):

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,594 @@
...
+    return endTrace.buildWordLattice(lexicon_);
+}
+
+bool LexiconfreeLabelsyncBeamSearch::decodeStep() {
+    if (finishedSegment_) {
+        return false;
+    }
+    if (currentSearchStep_ >= maxLabelsPerTimestep_ * std::max(totalTimesteps_, 1ul)) {
+        warning() << "Terminated search due to reaching max number of labels";
+        return false;
```
larissakl, line 254:
```suggestion
        warning() << "Terminated search due to reaching max number of labels";
        finishedSegment_ = true;
        return false;
```

larissakl:
I have just noticed that you have not registered this algorithm as a `SearchTypeV2` in Search/Module yet. I would say that we should implement this directly when introducing a new search algorithm.

src/Search/Module.hh
```diff
@@ -17,6 +17,7 @@
 
 #include <Core/Configuration.hh>
 #include <Core/Singleton.hh>
+#include "LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.hh"
```
larissakl, line 20: Is this new include needed?

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.hh
```diff
@@ -0,0 +1,192 @@
+/** Copyright 2026 RWTH Aachen University. All rights reserved.
```
larissakl, line 1: Why 2026?

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,662 @@
...
+
+/*
+ * =======================
+ * === LabelHypothesis ===
+ * =======================
+ */
+
+LexiconfreeLabelsyncBeamSearch::LabelHypothesis::LabelHypothesis()
+        : scoringContext(),
+          currentToken(Core::Type<Nn::LabelIndex>::max),
```
larissakl, line 37:
```suggestion
          currentToken(Nn::invalidLabelIndex),
```

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,662 @@
...
+            clog() << Core::XmlFull("best-active-hyp-score", bestActiveHyp->score);
+            clog() << Core::XmlFull("best-active-hyp-normalized-score", bestActiveHyp->scaledScore);
+        }
+        if (worstActiveHyp != nullptr) {
+            clog() << Core::XmlFull("worst-active-hyp-score", worstActiveHyp->score);
+            clog() << Core::XmlFull("worst-active-hyp-normalized-score", worstActiveHyp->scaledScore);
+        }
+        clog() << Core::XmlClose("search-step-stats");
+    }
+
```
larissakl, line 468: Why did you delete `++currentSearchStep_;` here?

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -68,6 +68,8 @@ LexiconfreeTimesyncBeamSearch::LabelHypothesis::LabelHypothesis(
             trace->score.acoustic = extension.score;
             trace->time           = extension.timeframe + 1;
             break;
+        default:
+            defect();  // Unexpected transition type which can not be produced by `inferTransitionType`
```
curufinwe: I am not a big fan of this. Crashing the search in case of boundary conditions is not appreciated when running workloads in a production setting. Maybe we could flag a warning and try to continue?

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,724 @@
...
+    labelScorer_ = modelCombination.labelScorer();
+
+    auto sentenceEndLemma = lexicon_->specialLemma("sentence-end");
+    if (!sentenceEndLemma) {
+        sentenceEndLemma = lexicon_->specialLemma("sentence-boundary");
+    }
+    if (sentenceEndLemma) {
+        if (sentenceEndLabelIndex_ == Core::Type<s32>::max) {
+            sentenceEndLabelIndex_ = sentenceEndLemma->id();
+            log() << "Use sentence-end index " << sentenceEndLabelIndex_ << " inferred from lexicon";
```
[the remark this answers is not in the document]
larissakl: Yes, but why should be assume that? I just had the problem that all the special lemmas in my lexicon were at the very end, and as result `sentenceEndLabelIndex_` became 200000. I feel like this very error prone as the user typically assumes that the order of the lemmas in the lexicon doesn't matter.

hannah220 (changes requested):

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,725 @@
...
+        : scoringContext(newScoringContext),
+          currentToken(extension.nextToken),
+          score(extension.score),
+          trace(),
+          isActive(extension.transitionType != Nn::LabelScorer::TransitionType::SENTENCE_END) {
+    switch (extension.transitionType) {
+        case Nn::LabelScorer::TransitionType::LABEL_TO_LABEL:
+        case Nn::LabelScorer::TransitionType::BLANK_TO_LABEL:
+        case Nn::LabelScorer::TransitionType::INITIAL_LABEL:
+        case Nn::LabelScorer::TransitionType::SENTENCE_END:
```
hannah220: Should SENTENCE_END be counted to the length?

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,725 @@
...
+    currentSearchStep_ = 0ul;
+
+    initializationTime_.stop();
+}
+
+void LexiconfreeLabelsyncBeamSearch::enterSegment(Bliss::SpeechSegment const* segment) {
+    initializationTime_.start();
+    labelScorer_->reset();
+    resetStatistics();
+    initializationTime_.stop();
```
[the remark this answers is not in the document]
SimBe195: This comes down to preference. The other searches also don't clean up the beam in `enterSegment` so if anything, it should be changed for all of them.

PR #128, opened 2025-05-27
Add Lm::OnnxStatelessLm

SimBe195:
This adds a new LM class `OnnxStatelessLm`.  This is a simple ONNX-based Language Model without any state caching. The entire token history is fed into the ONNX model for each score request. This trades efficiency for simplicity and flexibility. Thus, it is mostly useful for prototyping and models with a relatively small search space.


larissakl (changes requested):

larissakl (approved):

curufinwe (approved):
LGTM

PR #127, opened 2025-05-17
Add AedTreeBuilder

larissakl:
Adds an `AedTreeBuilder` which can later be used for a `TreeLabelsyncBeamSearch`. The search tree only includes labels, without blanks, self-loops and skip-transitions. 
The sentence-end token is retrieved from a special lemma in the lexicon and added as a label reachable from the root.
As in the `CtcTreeBuilder`, a word-boundary root will be added if a word-boundary token is present in the lexicon.
I moved some helper functions from `CtcTreeBuilder` to `AbstractTreeBuilder` so that I can easily reuse them.

PR #113, opened 2025-04-07
Add TreeTimesyncBeamSearch

larissakl:
Simple time synchronous beam search algorithm on a search tree built by the `CtcTreeBuilder` oder `RnaTreeBuilder`. The general structure is derived from the `LexiconfreeTimesyncBeamSearch`.

At a word end, a language model score is added to the hypothesis score. Supports global or separate pruning of within-word and word-end hypotheses by max beam-size and by score difference to the best hypothesis.

To simplify things and avoid over-complicated case distinctions, I introduced one minor adjustment in the TreeBuilder. The blank lemma does not have a self-loop anymore, when blank was predicted (as a "word") we can only go back to the root. I made sure that the AdvancedTreeSearch with the CTC tree is not affected and works as before.

PR #129, opened 2025-05-30
Add tree label-synchronous beam-search algorithm

larissakl:
This adds a label-synchronous search algorithm on a search tree built by the AedTreeBuilder (#127). 
It is derived from the lexicon-free label-synchronous beam-search algorithm (#126). Similar to the time-synchronous treesearch (#113), global or separate pruning of within-word and word-end hypotheses (within the set of active hypotheses) is possible and a language model score is added a word end.


src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,897 @@
...
+        case Nn::LabelScorer::TransitionType::INITIAL_LABEL:
+        case Nn::LabelScorer::TransitionType::SENTENCE_END:
+            // Start a new trace for the first word or for the sentence-end symbol
+            trace = Core::ref(new LatticeTrace(
+                    base.trace,
+                    extension.pron,
+                    extension.timeframe + 1,
+                    {extension.score - extension.lmScore, extension.lmScore},
+                    {}));
+            break;
```
hannah220: Empty line between `case` is maybe better to remove

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,897 @@
...
+                trace                 = Core::ref(new LatticeTrace(*base.trace));
+                trace->sibling        = {};
+                trace->pronunciation  = extension.pron;
+                trace->time           = extension.timeframe + 1;
+                trace->score.acoustic = base.trace->score.acoustic + (extension.score - base.score - extension.lmScore);
+                trace->score.lm       = base.trace->score.lm + extension.lmScore;
+            }
+            break;
+
+        default:
```
hannah220: What about LABEL_LOOP?
And in case there is blank, BLANK_LOOP, etc.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,897 @@
...
+    struct RecombinationContextHash {
+        size_t operator()(const RecombinationContext& context) const {
+            size_t h1 = context.state;
+            size_t h2 = Nn::ScoringContextHash{}(context.scoringContext);
+            size_t h3 = Lm::History::Hash{}(context.lmHistory);
+            return h1 ^ (h2 << 1) ^ (h3 << 2);
+        }
+    };
+
+    recombinedHypotheses_.clear();
```
hannah220:
```
    // Reserve capacity because future reallocations would break the raw pointer we are storing later
    recombinedHypotheses_.reserve(newBeam_.size());
```

hannah220 (changes requested):

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1248 @@
...
+
+const Core::ParameterBool TreeLabelsyncBeamSearch::paramLogStepwiseStatistics(
+        "log-stepwise-statistics",
+        "Log statistics about the beam at every search step.",
+        false);
+
+const Core::ParameterInt TreeLabelsyncBeamSearch::paramCacheCleanupInterval(
+        "cache-cleanup-interval",
+        "Interval of search steps after which buffered inputs that are not needed anymore get cleaned up.",
+        10);
```
curufinwe: `cache-cleanup-interval` has no lower bound here, so `0` is accepted and then used as a modulus (`currentSearchStep_ % cacheCleanupInterval_`) -> division by zero. Both `TreeTimesyncBeamSearch` and `LexiconfreeLabelsyncBeamSearch` pass a min of `1`.

```suggestion
        10,
        1);
```

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1248 @@
...
+          scoringContexts_(),
+          tempHypotheses_(),
+          currentSearchStep_(0ul),
+          totalTimesteps_(0ul),
+          finishedSegment_(false),
+          initializationTime_(),
+          featureProcessingTime_(),
+          scoringTime_(),
+          numHypsAfterIntermediatePruning_(),
+          numTerminatedHypsAfterScorePruning_("num-termianted-hyps-after-score-pruning"),
```
curufinwe: Typo in statistics label: `termianted` -> `terminated`.

```suggestion
          numTerminatedHypsAfterScorePruning_("num-terminated-hyps-after-score-pruning"),
```

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1248 @@
...
+     * We build a list of all scoring contexts that need to be passed to the LabelScorer for scoring scored inside `scoringContexts_`.
+     * `hypIndexToContextIndexMap_` stores the mapping, i.e. beam_[i].scoringContext = scoringContexts_[hypIndexToScoringContextMap_[i]].
+     * In the first iteration, this is just an identity mapping, i.e. hypIndexToContextIndexMap_[i] = i but for later label scorers
+     * some scoring contexts become no longer relevant when all extensions using them have been pruned.
+     */
+    withinWordExtensions_.clear();
+    scoringContexts_.clear();
+    scoringContexts_.reserve(beam_.size());
+    hypIndexToContextIndexMap_.resize(beam_.size());
+    std::iota(hypIndexToContextIndexMap_.begin(), hypIndexToContextIndexMap_.end(), 0ul);
```
curufinwe: Seeding the whole map with `iota` leaves stale indices for inactive hypotheses (harmless today since every read is guarded, but fragile). Matching `LexiconfreeLabelsyncBeamSearch`, assign `-1` and let the active-hyp loop below fill in the real indices.

```suggestion
    hypIndexToContextIndexMap_.assign(beam_.size(), -1);
```

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1248 @@
...
+                scoreTimes[accessorIdx]      = (*scoreAccessors[accessorIdx])->getTime();
+            }
+        }
+
+        if (scorerIdx == 0ul) {
+            // In the first iteration, create extensions while pre-pruning
+            Score currentBestScore = Core::Type<Score>::max;
+
+            for (size_t hypIndex = 0ul; hypIndex < beam_.size(); ++hypIndex) {
+                auto const hyp = beam_[hypIndex];
```
curufinwe: This copies the whole `LabelHypothesis` (scoring-context vector, LM history, trace ref) for every hypothesis each step; the loop only reads it. Use a const reference as in `LexiconfreeLabelsyncBeamSearch`.

```suggestion
                auto const& hyp = beam_[hypIndex];
```

curufinwe (commented):
A few small correctness/style fixes from reviewing against `TreeTimesyncBeamSearch` and `LexiconfreeLabelsyncBeamSearch`. Suggestions inline.
PS: 4 out of 6 Claude code suggestions. Have to verify the other two.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.hh
```diff
@@ -0,0 +1,286 @@
...
+    static const Core::ParameterBool        paramLogStepwiseStatistics;
+    static const Core::ParameterInt         paramMaximumStableDelay;
+    static const Core::ParameterInt         paramMaximumStableDelayPruningInterval;
+
+    TreeLabelsyncBeamSearch(Core::Configuration const&);
+
+    // Inherited methods from `SearchAlgorithmV2`
+
+    Speech::ModelCombination::Mode  requiredModelCombination() const override;
+    Speech::ModelCombination::Mode  requiredAcousticModel() const override;
```
curufinwe: Return type should be `Am::AcousticModel::Mode`, matching the base virtual (`SearchV2.hh:60`) and `TreeTimesyncBeamSearch`. It only compiles because both `Mode`s are typedefs for the same integer; the returned value (`Am::AcousticModel::noEmissions`) is already correct.

```suggestion
    Am::AcousticModel::Mode         requiredAcousticModel() const override;
```

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1248 @@
...
+    for (size_t i = 1ul; i <= maxBeamSizes_.size(); ++i) {
+        numHypsAfterIntermediatePruning_.push_back({"num-hyps-after-intermediate-pruning-" + std::to_string(i)});
+    }
+}
+
+Speech::ModelCombination::Mode TreeLabelsyncBeamSearch::requiredModelCombination() const {
+    return Speech::ModelCombination::useLabelScorer | Speech::ModelCombination::useLexicon | Speech::ModelCombination::useAcousticModel | Speech::ModelCombination::useLanguageModel;
+}
+
+Speech::ModelCombination::Mode TreeLabelsyncBeamSearch::requiredAcousticModel() const {
```
curufinwe: Match the corrected declaration type.

```suggestion
Am::AcousticModel::Mode TreeLabelsyncBeamSearch::requiredAcousticModel() const {
```

curufinwe (commented):
`requiredAcousticModel()` return type: use `Am::AcousticModel::Mode` like the base virtual and `TreeTimesyncBeamSearch` (type-only fix, no behavioral change).

curufinwe:
Offline discussion:
We want to refactor the code s.t. sentence-end is handled as a word-end hypothesis and creates a new trace. This would fix the following issue detected by Claude. In addition having the sentence-end score added before the end of the segment will change pruning behavior because the score for the terminated hypotheses is a bit higher now.

Here is the original Claude finding:
```
  What the code does

  Traces in this search are a word-level DAG, and node scores are cumulative from the start of the sentence. Two facts set up the bug:

  1. Within-word extensions share the base's trace node — TreeLabelsyncBeamSearch.cc:74, the within-word constructor does trace(base.trace). A new LatticeTrace is only ever created at a word end (the word-end constructor). So every in-word continuation of a word-end
  hypothesis points at the same LatticeTrace object, and the sentence-end token is itself a within-word transition (:75, isActive = transitionType != SENTENCE_END) — so a terminated hypothesis also keeps sharing its last word-end node.
  2. finalizeHypotheses mutates that shared node in place — :1172-1173 (and identically in the fallback branch :1186-1187):
  hyp.score          += sentenceEndScore;   // per-hypothesis — fine
  hyp.trace->score.lm += sentenceEndScore;   // mutates a SHARED LatticeTrace node

  Why it's wrong

  A single LatticeTrace node T (created at some word boundary) is referenced by every hypothesis that passed through that boundary and hasn't yet crossed the next one. At segment end the kept set is {terminated hyps} ∪ {active hyps at a root state}, and T is shared in
  two ways:

  - Double counting. Two terminated hypotheses that diverged after their last common word T (different token paths, so recombination doesn't merge them) both still have trace == T. Each executes T->score.lm += sentenceEndScore, so T.lm is inflated by 2× the score.
  - Cross-hypothesis corruption via predecessors. Because arc scores are node.score − predecessor.score, mutating T.lm shifts the arc into T by +delta and every arc out of T by −delta. Any other finalized hypothesis whose word chain runs through T as a predecessor
  (the common case — a word T that has both a continuation ending in silence/sentence-end and a continuation still active at the end) now has corrupted arc scores on its lattice path.

  Scope: this only corrupts trace/lattice scores (getCurrentBestWordLattice, getCurrentBestLatticeTrace, the score field in getCurrentBestTraceback). The 1-best word sequence is safe, because hypothesis selection uses hyp.score/scaledScore, not the trace scores.

  For contrast, TreeTimesyncBeamSearch::finalizeHypotheses never mutates existing nodes — it builds a new sentence-end node via the word-end constructor (TreeTimesyncBeamSearch.cc:1128-1156), so the sentence-end contribution lands on its own fresh, unshared arc.

  How I'd fix it

  Recommended — attach the sentence-end score on a fresh successor node, leaving all shared nodes (and their recombination sibling chains) untouched. Replace both mutation sites:

  if (not hyp.isActive or network_->isRoot(hyp.currentState)) {
      Lm::Score sentenceEndScore = languageModel_->sentenceEndScore(hyp.lmHistory);
      hyp.score += sentenceEndScore;
      // Put the sentence-end LM score on a fresh successor node instead of mutating the
      // (potentially shared) current trace node. Node scores are cumulative, so carry the
      // running totals forward with only the LM incremented.
      hyp.trace = Core::ref(new LatticeTrace(
              hyp.trace,                                                   // predecessor: left untouched
              nullptr,                                                     // no word emitted at sentence end
              hyp.trace->time,
              {hyp.trace->score.acoustic, hyp.trace->score.lm + sentenceEndScore},
              {}));
      newBeam_.push_back(hyp);
  }

  This mirrors the timesync design, is safe regardless of sharing, and the extra nullptr-pronunciation node is invisible to word traceback (both performTraceback word-listing and getCurrentBestWordLattice's endTrace already sit on top of hyp.trace and read its
  cumulative score). The one visible change vs. today is that the sentence-end LM becomes its own final ε-arc rather than being folded into the last word's arc — which is arguably more correct and matches timesync.

  Alternative — fix it at the source: make the SENTENCE_END within-word extension create its own trace node instead of trace(base.trace) (so terminated hyps own a distinct leaf). Then the in-place mutation becomes safe too. This is closer to how
  LexiconfreeLabelsyncBeamSearch gives every extension its own node, but it's more invasive (changes what a terminated hyp's trace/score represent) and I'd only take it if you want SENTENCE_END modeled as a real node throughout, not just at finalize.

  The design question for you: should the sentence-end be its own lattice arc (the fix above) or genuinely folded into the last word's arc? If the latter, there's no way to do it per-hypothesis without mutating a shared node — the score fundamentally belongs on a
  shared node — so the fresh-node approach is the honest resolution.

  Want me to apply the recommended fix (both branches) and post it as a suggestion, or hold for your call on the arc-vs-fold question?
  ```

**Comment by larissakl** [comment]
Now that I'm thinking about this, do we actually need to add the sentence-end **AM** scores in `finalizeHypotheses()` to the non-terminated hypotheses as well?
@curufinwe @SimBe195

**Comment by curufinwe** [comment]
The AM usually does not score sentence-end.

**Comment by larissakl** [comment]
Why not? In the TreeTimesyncBeamSearch we also let all label scorers score sentence end for each hypothesis in `finalizeHypotheses()`

**Comment by curufinwe** [comment]
For CTC models for example we do not usually include a sentence-end token in the softmax (or at least we don't train it). For RNN-T I guess it would make more sense to have this as part of the label set.

**Comment by larissakl** [comment]
But for this search, sentence-end is required, it's the stopping criterion for a hypothesis

**Comment by curufinwe** [comment]
That is true.

**Review by curufinwe — APPROVED** [review]

File: src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc [hunk 10]
```diff
@@ -0,0 +1,1367 @@
...
+            base.trace,
+            extension.pron,
+            timeframe + 1,
+            {totalAmScore, totalLmScore},
+            {}));
+}
+
+std::string TreeLabelsyncBeamSearch::LabelHypothesis::toString() const {
+    std::stringstream ss;
+    ss << "Score: " << score << ", traceback: ";
```
SimBe195: Maybe print scaled score in addition to full score.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1367 @@
+/** Copyright 2025 RWTH Aachen University. All rights reserved.
```
SimBe195:
```suggestion
/** Copyright 2026 RWTH Aachen University. All rights reserved.
```

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.hh
```diff
@@ -0,0 +1,292 @@
+/** Copyright 2025 RWTH Aachen University. All rights reserved.
```
SimBe195:
```suggestion
/** Copyright 2026 RWTH Aachen University. All rights reserved.
```

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1367 @@
...
+const Core::ParameterInt TreeLabelsyncBeamSearch::paramCacheCleanupInterval(
+        "cache-cleanup-interval",
+        "Interval of search steps after which buffered inputs that are not needed anymore get cleaned up.",
+        10,
+        1);
+
+const Core::ParameterInt TreeLabelsyncBeamSearch::paramMaximumStableDelay(
+        "maximum-stable-delay",
+        "Introduce a cutoff point at `current-time` - `delay`. Every hypothesis that disagrees with the current best anywhere before the cutoff gets pruned."
+        "This way words in the traceback become stable after at most `delay` frames.",
```
SimBe195: This concept doesn't really apply for label-sync search. The delay works in terms of emitted labels, not time. We should at least change the description but maybe also rename the parameter.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1367 @@
...
+        Am::AllophoneStateIndex alloStateIdx = acousticModel_->allophoneStateAlphabet()->index(&allo, 0);
+
+        sentenceEndLabelIndex_ = acousticModel_->emissionIndex(alloStateIdx);
+        log() << "Use sentence-end index " << sentenceEndLabelIndex_ << " inferred from lexicon";
+    }
+    else {
+        error() << "No sentence end lemma or pronunciation defined";
+    }
+
+    for (const auto& lemma : {"silence"}) {
```
SimBe195: Unnecessary loop over one item.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1367 @@
...
+                        .score          = hyp.score + languageModel_->sentenceEndScore(hyp.lmHistory),
+                        .timeframe      = hyp.timeframe,
+                        .transitionType = Nn::TransitionType::SENTENCE_END,
+                        .baseHypIndex   = hypIndex,
+                });
+                continue;
+            }
+
+            Score                               lmScore = 0;
+            const Bliss::SyntacticTokenSequence sts     = lemma->syntacticTokenSequence();
```
SimBe195:
```suggestion
            auto const& sts     = lemma->syntacticTokenSequence();
```

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1367 @@
...
+        }
+    }
+
+    /*
+     * Prune set of word-end hypotheses by max beam size and possibly also by score.
+     */
+    scorePruning(wordEndExtensions_, wordEndScoreThreshold_, maxWordEndBeamSize_);
+    numActiveWordEndHypsAfterIntermediatePruning_ += wordEndExtensions_.size();
+    if (logStepwiseStatistics_) {
+        clog() << Core::XmlFull("num-word-end-hyps-after-intermediate-pruning", wordEndExtensions_.size());
```
SimBe195: Why "intermediate"? In general, the order of pruning steps is a bit complicated in this search so ideally it should be documented somewhere.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1367 @@
...
+            std::remove_if(
+                    newBeam_.begin(),
+                    newBeam_.end(),
+                    [this](LabelHypothesis const& hyp) { return not hyp.isActive and not network_->isRoot(hyp.currentState); }),
+            newBeam_.end());
+
+    newBeam_.insert(newBeam_.end(), wordEndHypotheses_.begin(), wordEndHypotheses_.end());
+
+    /*
+     * Jointly prune terminated and active hypotheses by score
```
SimBe195, line 776: Within-word and word-end hypotheses are collected into `newBeam_` and then jointly pruned. This kind of defeats the purpose of the previous separation which I don't really like. But I see that not pruning globally means that we almost always retain within-word hyps (which are active by definition) which makes it difficult to reach the stop criterion. Perhaps in the future we can introduce different pruning modes and/or alternative stop criteria but doesn't need to be in this PR.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1367 @@
...
+        // Neither relative score pruning nor max beam size pruning triggers
+        return;
+    }
+
+    // Find ranges for score histogram and setting absolute threshold
+    Score lowerScore = Core::Type<Score>::max;
+    Score upperScore = Core::Type<Score>::min;
+
+    for (auto const& hyp : hypotheses) {
+        lowerScore = std::min(lowerScore, hyp.pruningScore());
```
SimBe195, line 998: Please protect against non-finite scores (see #247).

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1367 @@
...
+    withinWordExtensions_.clear();
+    for (size_t hypIndex = 0ul; hypIndex < tempHypotheses_.size(); ++hypIndex) {
+        auto const& hyp = tempHypotheses_[hypIndex];
+        withinWordExtensions_.push_back(
+                {sentenceEndLabelIndex_,
+                 hyp.currentState,
+                 hyp.trace->time,
+                 hyp.score,
+                 Nn::TransitionType::SENTENCE_END,
+                 hypIndex});
```
SimBe195: Prefer designated initializers.

SimBe195 (changes requested):

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1367 @@
...
+
+        wordEndHypotheses_.push_back({baseHyp, extension, newLmHistory, lengthNormScale_});
+    }
+
+    // Freshly terminated hypotheses have been expanded to word-end hypotheses, so the base hypotheses are removed
+    newBeam_.erase(
+            std::remove_if(
+                    newBeam_.begin(),
+                    newBeam_.end(),
+                    [this](LabelHypothesis const& hyp) { return not hyp.isActive and not network_->isRoot(hyp.currentState); }),
```
hannah220, line 770: Looking at the hypotheses that were skipped at line 684:
```
        // Terminated hypotheses that are already in a root state have been expanded to word-end
        // hypotheses at termination time and can't yield new ones
        if (not hyp.isActive and network_->isRoot(hyp.currentState)) {
```
Isn't the condition that you intend to prune: 
```suggestion
                    [this](LabelHypothesis const& hyp) { return not hyp.isActive and network_->isRoot(hyp.currentState); }),
```

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
larissakl: Hm ok, just for my understanding, the pruning cutoff is "currentSearchStep_ - delay" which we compare against `hyp.trace->time`, right? But `currentSearchStep_` only counts the decode steps, not the number of emitted labels, so the problem is that we are comparing a label-count cutoff against a frame-based trace time? Do you think a simple renaming/description rephrasing is enough here? Or should be maybe somehow adjust the general concept?

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -0,0 +1,1367 @@
...
+    size_t numActiveWordEnd = numActiveWordEndHyps();
+    numTerminatedHypsAfterRecombination_ += newBeam_.size() - numActive;
+    numActiveHypsAfterRecombination_ += numActive;
+    numActiveWordEndHypsAfterRecombination_ += numActiveWordEnd;
+    if (logStepwiseStatistics_) {
+        clog() << Core::XmlFull("num-terminated-hyps-after-recombination", newBeam_.size() - numActive);
+        clog() << Core::XmlFull("num-active-hyps-after-recombination", numActive);
+        clog() << Core::XmlFull("num-active-word-end-hyps-after-recombination", numActiveWordEnd);
+    }
+
```
hannah220, line 815:
```suggestion
/*
 * Prune new beam by max beam size
 */
```

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
SimBe195: The trace time is also label-based and not frame-based.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
larissakl: Maybe we can discuss the general pruning concept in this algorithm again offline?

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
larissakl: Oh, right, I see. So what about renaming it to `maximum-stable-label-delay`? Or do you have a better idea? Or we just leave the naming to keep it consistent with the timesync search and adjust the description?

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
larissakl: No, the goal here is to remove hypotheses which have just reached a word end in this step in order to replace them with a real word-end hypothesis (which also gets the sentence-end LM score etc.). These hypotheses are technically within-word hypotheses and therefore always at a non-root state, they are only redirected to the root state as soon as they become a word-end hypothesis. Therefore we need to remove all terminated within-word hypotheses, so those which are not active and which are not at a root state.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
larissakl: I called them "intermediate" here because it's not the final number after the "last" pruning step, so after the final joint pruning. I removed "intermediate" now, but I still find it confusing as we now have "num-word-end-hyps-after-pruning", "num-active-word-end-hyps-after-score-pruning" and "num-active-word-end-hyps-after-beam-pruning".

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
hannah220: Ok thanks!

hannah220 (approved):

SimBe195 (approved):

PR #130, opened 2025-06-02
Add interface to librasr for direct access to orthography

Stefanwuu:
In training, it's sometimes beneficial to directly access the orthography instead of only seeing the FSA. E.g. in sequence concatenation, it's easy to concatenate strings and then build FSA than concatenating FSAs directly. This was implemented by Daniel and I just want to make the PR for him.

### DIFF
--- a/src/Python/AllophoneStateFsaBuilder.hh
+++ b/src/Python/AllophoneStateFsaBuilder.hh
@@ -15,7 +15,7 @@ public:
     AllophoneStateFsaBuilder(const Core::Configuration& c);
     ~AllophoneStateFsaBuilder() = default;
 
-    std::string getOrthography(const std::string& segmentName);
+    std::string getOrthographyBySegmentName(const std::string& segmentName);
 
     py::tuple buildBySegmentName(const std::string& segmentName);
 
--- a/src/Python/AllophoneStateFsaBuilder.hh
+++ b/src/Python/AllophoneStateFsaBuilder.hh
@@ -15,6 +15,8 @@ public:
     AllophoneStateFsaBuilder(const Core::Configuration& c);
     ~AllophoneStateFsaBuilder() = default;
 
+    std::string getOrthography(const std::string& segmentName);
+
     py::tuple buildBySegmentName(const std::string& segmentName);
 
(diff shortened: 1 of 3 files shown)


DanEnergetics (approved):

src/Python/AllophoneStateFsaBuilder.hh
```diff
@@ -15,6 +15,8 @@ public:
     AllophoneStateFsaBuilder(const Core::Configuration& c);
     ~AllophoneStateFsaBuilder() = default;
 
+    std::string getOrthography(const std::string& segmentName);
```
albertz, line 18:
```suggestion
    std::string getOrthographyBySegmentName(const std::string& segmentName);
```
?

src/Tools/LibRASR/PybindModule.cc
```diff
@@ -23,6 +23,8 @@ PYBIND11_MODULE(librasr, m) {
 
     py::class_<AllophoneStateFsaBuilder> pyFsaBuilder(m, "AllophoneStateFsaBuilder");
     pyFsaBuilder.def(py::init<const Core::Configuration&>());
+    pyFsaBuilder.def("get_orthography",
```
albertz, line 26:
```suggestion
    pyFsaBuilder.def("get_orthography_by_segment_name",
```

Stefanwuu:
> Btw, why do you manually edit the changes in, and then don't click on resolve here? You can simply click on "commit suggestion" here on GitHub, and it will do all for you.

Thanks, commit suggestion was grey and I just noted I should do it from the file tab:(

albertz:
Btw, sorry for the multiple iterations, I always missed sth (but it seems you also missed them).

Btw, it seems there is no test? A simple build test would have caught any such problem.

Stefanwuu:
> Btw, sorry for the multiple iterations, I always missed sth (but it seems you also missed them).
> 
> Btw, it seems there is no test? A simple build test would have caught any such problem.

I should be sorry:( I actually fixed everything in my initial commit, but I didn't notice I git add . in the wrong directory, which wasted us too much time...

PR #132, opened 2025-06-08
Integrate LM lookahead into new TreeTimesyncBeamSearch

larissakl:
Integrates the LM lookahead (from Search/LanguageModelLookahead) into the new TreeTimesyncBeamSearch.
The lookahead is ~~activated~~ deactivated  and sparse per default. Optionally, a separate LM can be used.


curufinwe:
I went through the LM-lookahead integration in detail and pushed fixes for the back-off handling in 7b83548 — the commit message lists them. **One finding I deliberately did not fix, because I don't think there's a locally-correct fix and the right answer is probably yours to choose.**

### Recombination compares lookahead-inflated scores

`recombination` keys on state, scoring contexts and LM history — `TreeTimesyncBeamSearch.cc:988-989`:

```cpp
// Represents a unique combination of StateId, ScoringContext and LmHistory
struct RecombinationContext {
    StateId                            state;
    std::vector<Nn::ScoringContextRef> scoringContexts;
    Lm::History                        lmHistory;
```

Nothing about the lookahead. And the ordering in `decodeStep` is:

```
674    recombination(newBeam_, false);
680    scorePruning(newBeam_, Core::Type<Score>::max, maxBeamSizes_[labelScorers_.size() - 1]);
695            hyp.score -= hyp.lookaheadScore;
```

So recombination runs while `score` still carries the lookahead offset added at `:599`, and keeps the minimum of the *inflated* scores.

Two hypotheses can collide on that key while holding different lookahead tables: same state, same scoring contexts, same `lmHistory` — so the same lookahead context in the single-LM case — but one may have backed off earlier in the word and be carrying a lower-order table plus accumulated back-off while the other still holds the full-order one. Same successor state, different table, different offset. Then:

| | true score | lookahead offset | inflated |
|---|---|---|---|
| A | 10 | 5 | 15 |
| B | 12 | 1 | 13 |

B wins on 13 < 15, `:695` subtracts B's own offset restoring 12, and A — with the better true score — is gone. The arithmetic stays self-consistent, so nothing looks wrong; the search just kept the worse hypothesis. With `separate-lookahead-lm` it widens, since then equal `lmHistory` doesn't even imply equal lookahead context.

This is introduced by this PR: before it, `score` was the true score at `:674` and recombination was exact.

### Why I didn't fix it

Both obvious repairs cost something real:

- **Add the lookahead state to the key.** They stop being recombined and both survive. But it spends beam slots on hypotheses identical for every purpose except a pruning heuristic, and `lookahead` is a pointer — the key would become sensitive to table *identity*, i.e. cache state rather than search state.
- **Subtract before recombining.** Then `:674` and `:680` both see true scores. But that defeats the offset's purpose at `:680`, where beam pruning is supposed to rank by prospective cost. It trades a rare bad recombination for uniformly worse pruning.

Worth noting how `AdvancedTreeSearch` sidesteps this: `StateHypothesis` keeps `score` and `prospect` as **separate fields** (`SearchSpaceHelpers.hh:86-87`). Pruning compares `prospect`, recombination and everything durable use `score`, and no subtraction is ever needed. This algorithm folds both into one `score`, which is what forces the add-then-subtract at `:599`/`:695` and produces this finding. A `prospect` field would dissolve it and remove the dance — but that's a design change, not a review fix, so I've left it to you.

Happy to be talked out of any of it. Low severity in isolation: a suboptimal recombination, not wrong output.

— Claude Code

curufinwe (approved):

SimBe195 (approved):

PR #137, opened 2025-07-23
Enable GPU support for ONNX sessions

SimBe195:
This PR enables ONNX session to run on GPU. At runtime, RASR detects whether a GPU is available and adds a `CUDAExecutionProvider` to the ONNX session if that is the case. To enable this behavior, RASR needs to be compiled with `MODULE_CUDA` since the `cuda_runtime.h` header is required to detect a GPU presence. If the `CUDAExecutionProvider` is added without a GPU, it will unfortunately result in a `segmentation fault` inside the ONNX runtime and I have not found another way to detect GPU presence with ONNX alone and no CUDA headers.

Depends on #136 for compilation with `MODULE_CUDA`.

### DIFF
--- a/src/Onnx/Session.cc
+++ b/src/Onnx/Session.cc
@@ -5,6 +5,10 @@
 
 #include <chrono>
 
+#ifdef MODULE_CUDA
+#include <cuda_runtime.h>
+#endif
+
 #include "Util.hh"
 
 namespace Onnx {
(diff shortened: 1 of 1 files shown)


larissakl (approved):

src/Onnx/Session.cc
```diff
@@ -34,6 +46,36 @@ Session::Session(Core::Configuration const& config)
...
+#endif
+            int deviceCount = 0;
+            if (cudaGetDeviceCount(&deviceCount) != cudaSuccess or deviceCount == 0) {
+                error() << "Requested CUDA execution provider but no CUDA device was found.";
+            }
+            OrtCUDAProviderOptionsV2* cuda_opts = nullptr;
+            Ort::ThrowOnError(Ort::GetApi().CreateCUDAProviderOptions(&cuda_opts));
+            session_opts.AppendExecutionProvider_CUDA_V2(*cuda_opts);
+            Ort::GetApi().ReleaseCUDAProviderOptions(cuda_opts);
+            break;
```
larissakl, line 73:
```suggestion
#ifdef MODULE_CUDA
            int deviceCount = 0;
            if (cudaGetDeviceCount(&deviceCount) != cudaSuccess or deviceCount == 0) {
                error() << "Requested CUDA execution provider but no CUDA device was found.";
            }
            OrtCUDAProviderOptionsV2* cuda_opts = nullptr;
            Ort::ThrowOnError(Ort::GetApi().CreateCUDAProviderOptions(&cuda_opts));
            session_opts.AppendExecutionProvider_CUDA_V2(*cuda_opts);
            Ort::GetApi().ReleaseCUDAProviderOptions(cuda_opts);
            break;
#endif
```
I had to add this because I ran into the errors "‘cudaGetDeviceCount’ was not declared in this scope" and 
"‘cudaSuccess’ was not declared in this scope" when trying to compile with MODULE_CUDA disabled.

curufinwe (approved):

PR #138, opened 2025-07-24
Add TransitionLabelScorer

SimBe195:
This PR adds a new label scorer `TransitionLabelScorer` which wraps a base LabelScorer and adds predefined transition scores to the base scores depending on the transition type of each request. The transition scores are all individually specified as config parameters.


larissakl (approved):

curufinwe:
I'm OK with the current implementation, but if you want I have some suggestions on how to make it more compact:
1. make the LabelScorer::TransitionType enum an explicit int (by making it enum class).
2. Then store an array of strings for the config names and create the parameters on the fly instead of having 7 different members.
3. Same for the scores.
4. The lookup of the transition score then becomes an array lookup without an explicit switch statement. Bonus points for adding a range check here.
 
Then when a new transition type is added we can just add another value to the string array and be done with it. Bonus points for adding a sentinel to the enum at the end and adding a compile time check to enforce this scorer having the right number of elements in the array.

curufinwe (approved):

PR #145, opened 2025-09-19
Modify TreeTimesyncBeamSearch: don't force LM to ignore blank and silence

larissakl:
In the TreeTimesyncBeamSearch, we had a hardcoded check to force the LM to ignore the blank and silence lemma. There are however use cases where blank or silence are part of the LM as well. Therefore, we removed this hardcoded check and only check if the syntactic token sequence is not empty.
This also means that in the special lemmas blank and silence, the user needs to set `<synt/>` in the lexicon, otherwise the LM will score these lemmas. As this can easily happen by mistake, we added a warning for that.

### DIFF
--- a/src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
+++ b/src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
@@ -212,6 +212,12 @@ bool TreeTimesyncBeamSearch::setModelCombination(Speech::ModelCombination const&
         useBlank_        = false;
     }
 
+    for (const auto& lemma : {"silence", "blank"}) {
+        if (lexicon_->specialLemma(lemma) and (lexicon_->specialLemma(lemma)->syntacticTokenSequence()).size() != 0) {
+            warning("Special lemma \"%s\" will be scored by the language model. To prevent the LM from scoring it, set an empty syntactic token sequence for it in the lexicon.", lemma);
+        }
+    }
+
     // Create look-ups for state successors and exits of each state
     createSuccessorLookups();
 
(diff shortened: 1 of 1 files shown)


src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
```diff
@@ -421,9 +427,10 @@ bool TreeTimesyncBeamSearch::decodeStep() {
                                                     Nn::LabelScorer::TransitionType::INITIAL_BLANK,  // The transition type is irrelevant, so just use this as dummy
                                                     hypIndex};
 
-                if (lemma != lexicon_->specialLemma("blank") and lemma != lexicon_->specialLemma("silence")) {
-                    const Bliss::SyntacticTokenSequence sts = lemma->syntacticTokenSequence();
-                    const Bliss::SyntacticToken*        st  = sts.front();
+                const Bliss::SyntacticTokenSequence sts = lemma->syntacticTokenSequence();
+                if (not(sts.size() == 0)) {
```
SimBe195, line 431:
```suggestion
                if (not sts.empty()) {
```

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
```diff
@@ -447,11 +454,12 @@ bool TreeTimesyncBeamSearch::decodeStep() {
...
     wordEndHypotheses_.clear();
     for (auto& extension : extensions_) {
-        const Bliss::Lemma* lemma = extension.pron->lemma();
-        if (lemma != lexicon_->specialLemma("blank") and lemma != lexicon_->specialLemma("silence")) {
-            const Bliss::SyntacticTokenSequence sts = lemma->syntacticTokenSequence();
-            const Bliss::SyntacticToken*        st  = sts.front();
-            extension.lmHistory                     = languageModel_->extendedHistory(extension.lmHistory, st);
+        const Bliss::Lemma*                 lemma = extension.pron->lemma();
+        const Bliss::SyntacticTokenSequence sts   = lemma->syntacticTokenSequence();
+        if (not(sts.size() == 0)) {
```
SimBe195, line 459:
```suggestion
        if (not sts.empty()) {
```

SimBe195 (approved):

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
larissakl: This is not possible as `SyntacticTokenSequence` is not a vector but a `SymbolSequence` (see Bliss/Symbol.hh) which does not have a `empty()` function. However, if you prefer, I can add this function to this class.

curufinwe (approved):

PR #146, opened 2025-09-23
Update Nn::StatefulOnnxLabelScorer to compute hidden state only once scores are requested

larissakl:
The function `finalizeScoringContext()` is added to the LabelScorer. The `StatefulOnnxLabelScorer` won't update the hidden state in `extendedScoringContext()` anymore, but only append the new token to the sequence. The hidden state update is now done in this new function.  The default implementation which is used in the other label scorers is just no-op. For future label scorers, one should pay attention to put costly operations which don't affect the hash into `finalizeScoringContext()` instead of `extendedScoringContext()` if possible.
In the search algorithms, the order is now:

1. Score Pruning
2. Extended Scoring Context
3. Recombination
4. Beam Size Pruning
5. Finalize Scoring Context

The last step is not done at the end of the decode step, but at the beginning of the next decode step such that we save unnecessary calculations at the end of a segment.

### DIFF
--- a/src/Nn/LabelScorer/ScoringContext.hh
+++ b/src/Nn/LabelScorer/ScoringContext.hh
@@ -139,8 +139,8 @@ struct OnnxHiddenStateScoringContext : public ScoringContext {
     OnnxHiddenStateScoringContext()
             : labelSeq(), hiddenState(), requiresFinalize(false) {}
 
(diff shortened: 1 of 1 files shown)


SimBe195:
I believe in our regular RASR meeting we decided to do this a bit differently and keep the logic only internally in the StatefulLabelScorer instead of modifying the interface or search algorithms directly. I can do the re-write on this branch accordingly but until then, the PR shouldn't be merged.

PR #148, opened 2025-10-01
Update Nn::LabelScorer: allow enabling of specific transition types

SimBe195:
~~This adds a parameter `ignored-transition-types`~~ to the LabelScorer base class which allows users to specify a list of transition types which will always get assigned score 0 and not affect the ScoringContext. This can be especially helpful when building combined label scorers with multiple sub-scorers of different label topology such as for example CTC + LSTM LM where the LM needs to ignore all blank transitions as well as label loops and CTC on the other hand needs to ignore the sentence-end.

Do achieve this, `extendedScoringContext`, `computeScoreWithTime` and `computeScoresWithTimes` are implemented in the base `LabelScorer` and call protected functions `{extendedScoringContext|computeScoreWithTime|computeScoresWithTimes}Internal` which are overridden in child classes.

Depends on modifications to the transition type handling from #138.

Edit:
This logic was changed so that instead of a blacklist, the LabelScorer has a parameter `transition-preset` which enables a preconfigured list of used transition types for specific models and any additional ones can be enabled via another parameter `extra-transition-types`. The available presets for now are `"default"`, `"none"`, `"ctc"`, `"transducer"` and `"lm"`. When the `default` preset is used, it takes a preset that is specified by each LabelScorer subclass individually and should be the most common use-case for that class. For example for `CombineLabelScorer` the default will use preset "all", for `NoContextOnnxLabelScorer` the default will use preset "ctc" and so on.


larissakl (approved):

larissakl (changes requested):

src/Nn/LabelScorer/LabelScorer.hh
```diff
@@ -155,6 +168,24 @@ protected:
...
+
+    virtual ScoringContextRef            extendedScoringContextInternal(Request const& request) = 0;
+    virtual std::optional<ScoreWithTime> computeScoreWithTimeInternal(Request const& request)   = 0;
+
+    // By default loops over the single-request version
+    virtual std::optional<ScoresWithTimes> computeScoresWithTimesInternal(std::vector<Request> const& requests);
+
+    virtual TransitionPresetType defaultPreset() const {
+        return TransitionPresetType::NONE;
+    }
```
curufinwe: It would be nice to move method definitions out of the class. Can either be inline in the header or just in the .cc file.

src/Nn/LabelScorer/StatefulOnnxLabelScorer.hh
```diff
@@ -74,18 +74,22 @@ public:
...
 
+    // Forward hidden-state through state-updater ONNX model
+    Core::Ref<const ScoringContext> extendedScoringContextInternal(LabelScorer::Request const& request) override;
+
+    std::optional<LabelScorer::ScoreWithTime>   computeScoreWithTimeInternal(LabelScorer::Request const& request) override;
+    std::optional<LabelScorer::ScoresWithTimes> computeScoresWithTimesInternal(std::vector<LabelScorer::Request> const& requests) override;
+
+    virtual TransitionPresetType defaultPreset() const override {
+        return TransitionPresetType::LM;
+    }
```
curufinwe: Move definition out of class declaration.

src/Nn/LabelScorer/NoContextOnnxLabelScorer.hh
```diff
@@ -44,32 +44,36 @@ public:
...
 
     // Uses `getScoresWithTimes` internally with some wrapping for vector packing/expansion
-    std::optional<LabelScorer::ScoreWithTime> computeScoreWithTime(LabelScorer::Request const& request) override;
+    std::optional<LabelScorer::ScoreWithTime> computeScoreWithTimeInternal(LabelScorer::Request const& request) override;
 
-protected:
-    size_t getMinActiveInputIndex(Core::CollapsedVector<ScoringContextRef> const& activeContexts) const override;
+    virtual TransitionPresetType defaultPreset() const override {
+        return TransitionPresetType::CTC;
+    }
```
curufinwe: see others

curufinwe (changes requested):

src/Nn/LabelScorer/LabelScorer.cc
```diff
@@ -89,12 +92,13 @@ std::optional<LabelScorer::ScoresWithTimes> LabelScorer::computeScoresWithTimes(
     }
 
     // Interleave actual results with 0 scores for requests with ignored transition types
-    ScoresWithTimes result{{requests.size(), 0.0}, {requests.size(), 0}};
+    ScoresWithTimes result{
+            .scores = std::vector<Score>(requests.size(), 0.0),
+            .timeframes{requests.size(), 0}};
```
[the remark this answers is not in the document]
SimBe195: clang-format complains about putting the `};` in a new line.

curufinwe (approved):

PR #149, opened 2025-10-02
Add stateful-transducer-onnx label scorer

SimBe195:
Add a new label scorer class `StatefulTransducerOnnxLabelScorer`. This is similar to the `StatefulOnnxLabelScorer` but hidden-state and encoder features are decoupled. Here, the state initializer and state updater only take tokens and no encoder features. Instead, the scorer receives the encoder state at the current time step and the hidden state and is supposed to combine the two.

This can be used for transducer models with unlimited context.

Depends on #144.


src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.cc
```diff
@@ -0,0 +1,425 @@
...
+    std::vector<LabelIndex> newLabelSeq(scoringContext->labelSeq);
+    bool                    requiresFinalize = false;
+
+    if (pushToken) {
+        newLabelSeq.push_back(request.nextToken);
+        requiresFinalize = true;
+    }
+
+    // Re-use previous hidden-state but mark that finalization (i.e. hidden-state update) is required
+    auto newScoringContext              = Core::ref(new StepOnnxHiddenStateScoringContext(scoringContext->currentStep + timeIncrement, std::move(newLabelSeq), scoringContext->hiddenState));
```
hannah220: You could directly add `requiresFinalize` in the constructor above.

src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.hh
```diff
@@ -0,0 +1,137 @@
...
+    bool   verticalLabelTransition_;
+    size_t maxBatchSize_;
+
+    Onnx::Model scorerOnnxModel_;
+    Onnx::Model stateInitializerOnnxModel_;
+    Onnx::Model stateUpdaterOnnxModel_;
+
+    StepOnnxHiddenStateScoringContextRef initialScoringContext_;
+
+    // Map input/output names of onnx models to hidden state names taken from state initializer model
```
larissakl: I don't get this comment. For example the `updaterInputToStateNameMap_` is not taken from the state initializer model, is it? You only check for consistency with the names of the initializer.

src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.hh
```diff
@@ -0,0 +1,137 @@
...
+
+    std::string updaterTokenName_;
+
+    Core::FIFOCache<StepOnnxHiddenStateScoringContextRef, std::shared_ptr<std::vector<Score>>, ScoringContextHash, ScoringContextEq> scoreCache_;
+    Core::FIFOCache<StepOnnxHiddenStateScoringContextRef, OnnxHiddenStateRef, ScoringContextHash, ScoringContextEq>                  stateCache_;
+};
+
+}  // namespace Nn
+
+#endif  // STATEFUL_ONNX_LABEL_SCORER_HH
```
larissakl:
```suggestion
#endif  // STATEFUL_TRANSDUCER_ONNX_LABEL_SCORER_HH
```

src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.cc
```diff
@@ -0,0 +1,501 @@
...
+
+#include "ScoreAccessor.hh"
+#include "ScoringContext.hh"
+
+namespace Nn {
+
+/*
+ * =============================
+ * == StatefulTransducerOnnxLabelScorer ==
+ * =============================
```
larissakl:
```suggestion
 * =======================================
 * == StatefulTransducerOnnxLabelScorer ==
 * =======================================
```
Sorry but this triggers me :smile:

src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.cc
```diff
@@ -0,0 +1,501 @@
...
+
+const Core::ParameterInt StatefulTransducerOnnxLabelScorer::paramMaxBatchSize(
+        "max-batch-size",
+        "Max number of hidden-states that can be fed into the scorer ONNX model at once.",
+        Core::Type<int>::max);
+
+const Core::ParameterInt StatefulTransducerOnnxLabelScorer::paramMaxCachedScores(
+        "max-cached-score-vectors",
+        "Maximum size of cache that maps histories to scores. This prevents memory overflow in case of very long audio segments.",
+        1000);
```
larissakl: I know that in the `StatefulOnnxLabelScorer`, we also have 1000 as default for `paramMaxCachedScores`, but I think we should consider increasing it for both label scorers, it seems a bit too low to me.

src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.hh
```diff
@@ -0,0 +1,137 @@
...
+    void cacheScores(std::vector<StepOnnxHiddenStateScoringContextRef> const& scoringContextBatch);
+
+    // Computes new hidden state based on previous hidden state and next token with batched state-updater call
+    std::vector<OnnxHiddenStateRef> updatedHiddenStates(std::vector<OnnxHiddenStateRef> const& hiddenStatesBatch, std::vector<s32> nextTokensBatch);
+
+    // Compute updated states for all non-finalized scoring contexts and put them into the state cache
+    void cacheStates(std::vector<StepOnnxHiddenStateScoringContextRef> const& scoringContextBatch);
+
+    // Replace hidden-state in scoringContext with an updated version that includes the last label
+    void finalizeScoringContext(StepOnnxHiddenStateScoringContextRef const& scoringContext);
```
larissakl: I think this function is not needed anymore as the hidden state updating is done in `cacheStates()`

src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.cc
```diff
@@ -0,0 +1,501 @@
...
+void StatefulTransducerOnnxLabelScorer::cacheStates(std::vector<StepOnnxHiddenStateScoringContextRef> const& scoringContextBatch) {
+    std::vector<StepOnnxHiddenStateScoringContextRef> nonFinalizedContexts;
+    std::vector<OnnxHiddenStateRef>                   hiddenStates;
+    std::vector<s32>                                  nextTokens;
+    for (auto const& scoringContext : scoringContextBatch) {
+        if (not scoringContext->requiresFinalize) {
+            continue;
+        }
+        nonFinalizedContexts.push_back(scoringContext);
+        verify(scoringContext->hiddenState);
```
larissakl, line 407: In the StatefulOnnxLabelScorer, we have this:
```
        if (scoringContext->hiddenState) {
            hiddenStates.push_back(scoringContext->hiddenState);
        }
        else {
            hiddenStates.push_back(computeInitialHiddenState());
        }
```
Maybe this is actually not required? However I don't see a reason why it should be in there but not here. (same in `cacheScores()`)

src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.hh
SimBe195: It's written like this because in the code the initializer defines all the state names (since it has to cover every state) and the state names of the other models are always compared against the names taken from the initializer model. But perhaps this comment is a bit confusing so I will just reduce it to "Map input/output names of onnx models to hidden state names" (also in the `StatefulOnnxLabelScorer`.

src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.hh
SimBe195: Good catch, `finalizeScoringContext` is actually never called anymore.

src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.cc
SimBe195: Yes I agree, this often has to be overwritten in config. I'll set it to 10k for now.

src/Nn/LabelScorer/StatefulTransducerOnnxLabelScorer.cc
SimBe195: The reason for this is that the state initializer in the `StatefulOnnxLabelScorer` takes encoder-states as input, so it uses a sentinel empty hidden-state in order to delay the computation until all inputs have been passed. This is not the case here; states are independent of the encoder and the initial state can be properly computed immediately in `getInitialScoringContext`.

larissakl: Ah, I see, makes sense. Thanks for the explanation.

larissakl (approved):

PR #154, opened 2025-10-20
Add HmmTreeBuilder

larissakl:
Adds a treebuilder for the HMM-topology which is, in contrast to the old treebuilders ("minimized-hmm" and "classic-hmm"), compatible with the new TreeTimesyncBeamSearch.

For monophones, the tree is very simple:
tree_monophone_1_noctx

To get this tree, the variation of the phonemes in the lexicon needs to be set to "none". If it's set to "context", it will work as well, but have unnecessary root states:
tree_monophone_1

For diphones (diphone-dense state tying), the tree looks as follows:
tree_diphone_1

Triphones are not supported. Depends on #127 as I am using the shared base class here as well.

I did one test run with the TreeTimesyncBeamSearch and there are not obvious errors. However, I still observe a degradation from 5.8% to 6.2% which could have multiple reasons. Therefore, some considerations for future work:
In order to get competitive results, there might still be some details we need to implement. Most importantly, we need an exit penalty, which might be easy to add to the search, but we should probably allow for different exit penalties for words and non-words (silence/unknown). Furthermore, with the TransitionLabelScorer, it is easy to define TDPs, but again, we should have the possibility to have different TDPs for non-words, so we would need more transition types (especially silence-transitions) and tiny adjustments in the search algorithms.


src/Search/TreeBuilder.cc
```diff
@@ -1598,3 +1603,260 @@ void AedTreeBuilder::addWordBoundaryStates() {
...
+        : SharedBaseClassTreeBuilder(config, lexicon, acousticModel, network),
+          addCiTransitions_(paramAddCiTransitions(config)) {
+    const auto iters = lexicon.phonemeInventory()->phonemes();
+
+    if (initialize) {
+        verify(lexicon.specialLemma("silence") != nullptr);
+        verify(!network_.rootState);
+        network_.ciRootState = network_.rootState = createRoot();  // context-independent root
+
+        // network_.finalStates.insert(network_.rootState);
```
SimBe195: Why is this commented out?

src/Search/TreeBuilder.cc
```diff
@@ -1598,3 +1603,260 @@ void AedTreeBuilder::addWordBoundaryStates() {
...
+
+        for (auto it = iters.first; it != iters.second; ++it) {
+            const auto* phoneme = *it;
+
+            if (phoneme->isContextDependent()) {
+                // Create coarticulated root states for context dependent phonemes
+                const StateId root = createRoot();
+                network_.coarticulatedRootStates.insert(root);
+                // network_.finalStates.insert(root);
```
SimBe195: See above.

src/Search/TreeBuilder.cc
```diff
@@ -1598,3 +1603,260 @@ void AedTreeBuilder::addWordBoundaryStates() {
...
+                        addExit(lastState, network_.rootState, (*it)->id());
+                    }
+                }
+            }
+            // Special handling for lemmas with more than one phoneme
+            else {
+                if (acousticModel_.phonemeInventory()->phoneme(lastPhoneme)->isContextDependent() and lastPhoneme != Bliss::Phoneme::term) {
+                    connectRoot(contextRoot, contextPhoneme, pron);
+                }
+            }
```
SimBe195, line 1703: This nested if-else logic seems confusing to me and also doesn't seem to be handling all possible cases. For example if pronunciation length is > 1 and the last phoneme is not context dependent, nothing gets connected.

src/Search/TreeBuilder.cc
```diff
@@ -1598,3 +1603,260 @@ void AedTreeBuilder::addWordBoundaryStates() {
...
+                break;
+            }
+        }
+        else {
+            // Second phoneme generation: connect to existing states (if not already connected to first phoneme generation of context-independent root)
+            const auto desc  = hmm[0];
+            const auto state = secondGenPhonemes_[desc];
+            addTransition(currentState, state);
+        }
+    }
```
SimBe195, line 1860: This is again deeply nested logic which is difficult to follow. The control flow could be refactored a little bit e.g. via separating by generation or introducing helper functions.

SimBe195 (changes requested):

src/Search/TreeBuilder.hh
```diff
@@ -45,40 +45,6 @@ public:
...
-
-    MinimizedTreeBuilder(Core::Configuration config, const Bliss::Lexicon& lexicon, const Am::AcousticModel& acousticModel, Search::PersistentStateTree& network, bool initialize = true);
-    virtual ~MinimizedTreeBuilder() = default;
-
-    virtual std::unique_ptr<AbstractTreeBuilder> newInstance(Core::Configuration config, const Bliss::Lexicon& lexicon, const Am::AcousticModel& acousticModel, Search::PersistentStateTree& network, bool initialize = true);
-
-    virtual void build();
-
 protected:
     struct HMMSequence {
```
curufinwe: Instead of moving this into AbstractTreeBuilder I would suggest you move this class out of any TreeBuilder class s.t. anyone that needs it can use it.

curufinwe (changes requested):
I only had a look at the header and not the actual tree-builder logic. We can merge from my side even if the TreeBuilder itself is not 100% good. But if @SimBe195 and you @larissakl want to rework the logic it's fine with me.

curufinwe (approved):

SimBe195 (approved):

PR #179, opened 2026-02-05
Add Search::LexiconfreeRNNTTimesyncBeamSearch

larissakl:
Adds a lexiconfree timesynchronous beam-search algorithm for standard (non-monotonic) Transducers. At each timestep, multiple non-blank labels can be predicted (the maximum number is controllable via a hyperparameter), a hypothesis is finished in the current timestep if it has emitted a blank label. In the inner loop of a timestep, first, all active inner hypotheses are extended with blank so they become outer hypotheses. Then, the inner hyps are extended by non-blank tokens and are pruned. If there are already more than max-beam-size outer hyps, all inner hyps which are worse than the worst of the max-beam-size best outer hyps are removed. If no inner hyps are left, the inner loop is stopped. At the end of a timestep, the outer hyps are pruned again based on their length-normalized score. 

The implementation is based on [PyTorch's RNNTBeamSearch](https://docs.pytorch.org/audio/main/_modules/torchaudio/models/rnnt_decoder.html#RNNTBeamSearch)

Major To-Dos:
- Integration of sub-scorers and intermediate pruning
- Correct handling of sentence-end
- Testing


src/Search/LexiconfreeRNNTTimesyncBeamSearch/LexiconfreeRNNTTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,1052 @@
...
+        0.0);
+
+const Core::ParameterInt LexiconfreeRNNTTimesyncBeamSearch::paramMaxLabelsPerFrame(
+        "max-labels-per-timeframe",
+        "Maximum number of non-blank label predictions per hypothesis in one timestep.",
+        10, 0);
+
+const Core::ParameterInt LexiconfreeRNNTTimesyncBeamSearch::paramBlankLabelIndex(
+        "blank-label-index",
+        "Index of the blank label in the lexicon. Can also be inferred from lexicon if it has a lemma with `special='blank'`. If not set, the search will not use blank.",
```
hannah220: > If not set, the search will not use blank.

In this search, blank is not optional.

src/Search/LexiconfreeRNNTTimesyncBeamSearch/LexiconfreeRNNTTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,1052 @@
...
+
+            auto denseScores = (*scoreAccessor)->getDenseScores();
+            auto scoreTime   = (*scoreAccessor)->getTime();
+
+            // Iterate over possible successors (all lemmas)
+            for (auto lemmaIt = lemmas.first; lemmaIt != lemmas.second; ++lemmaIt) {
+                const Bliss::Lemma* lemma(*lemmaIt);
+                Nn::LabelIndex      tokenIdx = lemma->id();
+
+                // Blank is not allowed as an extension for the inner hyps
```
hannah220: Isn't the reason because extension with blank label for inner hyps is already done? (line 451~472)

src/Search/LexiconfreeRNNTTimesyncBeamSearch/LexiconfreeRNNTTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,1052 @@
...
+        // because there is only one blank candidate per inner hyp, so an intermediate cut isn't needed for performance here
+        scoreWithRemainingLabelScorers(extensions_, innerHyps_);
+
+        // Create new label hypotheses from extension candidates
+        newBeam_.clear();
+        for (auto const& extension : extensions_) {
+            auto const& baseHyp = innerHyps_[extension.baseHypIndex];
+            newBeam_.push_back({baseHyp, extension, extendedScoringContexts(baseHyp, extension), lengthNormScale_});
+        }
+
```
hannah220: The next time newBeam_ is used is at line 554 which is `newBeam_.clear();,` so do you need to insert to `newBeam_` before copying to `outerHyps_`?
Can you just insert to `outerHyps_` directly?

src/Search/LexiconfreeRNNTTimesyncBeamSearch/LexiconfreeRNNTTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,1052 @@
...
+        // Prune new inner hyps down to maxBeamSize based on the raw score
+        beamSizePruning(newBeam_);
+
+        // If there are already more than maxBeamSize outer hyps,
+        // remove all inner hyps with a score that is lower than the worst score of the max-beam-size best outer hyps
```
hannah220:
```suggestion
        // remove all inner hyps with a score that is greater than the worst score of the max-beam-size best outer hyps
```

src/Search/LexiconfreeRNNTTimesyncBeamSearch/LexiconfreeRNNTTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,1052 @@
...
+            if (not scoreAccessor) {
+                // Extension is not scorable so set the score to max in order to prune it later
+                ext.score = Core::Type<Score>::max;
+                continue;
+            }
+
+            auto const& denseScores = denseScoreSpans[contextIdx];
+            ext.score += (denseScores and ext.nextToken < denseScores->size())
+                                 ? (*denseScores)[ext.nextToken]
+                                 : (*scoreAccessor)->getScore(ext.transitionType, ext.nextToken);
```
hannah220: Does this guarantee that only at blank, the time stamp increment?

src/Search/LexiconfreeRNNTTimesyncBeamSearch/LexiconfreeRNNTTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,1052 @@
...
+            ext.timeframe = std::max(ext.timeframe, scoreTimes[contextIdx]);
+        }
+
+        size_t maxBeamSize = extensions.size();
+        if (scorerIdx < labelScorers_.size() - 1) {
+            maxBeamSize = maxBeamSizes_[scorerIdx];
+        }
```
hannah220: Maybe it's better to put in the function name that this also does pruning.

src/Search/LexiconfreeRNNTTimesyncBeamSearch/LexiconfreeRNNTTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,1052 @@
...
+    hypotheses.resize(maxBeamSizes_.back());  // Get rid of excessive elements
+}
+
+void LexiconfreeRNNTTimesyncBeamSearch::beamSizePruningLengthnormalized(std::vector<LabelHypothesis>& hypotheses) const {
+    if (hypotheses.size() <= maxBeamSizes_.back()) {
+        return;
+    }
+
+    // Reorder the hypotheses by associated scaledScore value such that the first `beamSize_` elements are the best
+    std::nth_element(hypotheses.begin(), hypotheses.begin() + maxBeamSizes_.back(), hypotheses.end());
```
hannah220: Don't you need `[](auto const& a, auto const& b) { return a.score < b.score; }` as compare function?
Also, length normalization does not play a role for beam size pruning, so we don't need this extra function?

src/Search/LexiconfreeRNNTTimesyncBeamSearch/LexiconfreeRNNTTimesyncBeamSearch.cc
```diff
@@ -0,0 +1,1052 @@
...
+
+void LexiconfreeRNNTTimesyncBeamSearch::scorePruningLengthnormalized(std::vector<LabelHypothesis>& hypotheses) const {
+    if (hypotheses.empty()) {
+        return;
+    }
+
+    // Compute the pruning threshold
+    auto bestHyp = *std::min_element(
+            hypotheses.begin(),
+            hypotheses.end());
```
hannah220: You also need `[](auto const& a, auto const& b) { return a.score < b.score; })->score;`?

PR #196, opened 2026-03-18
Add histogram pruning for SearchV2 algorithms

SimBe195:
This PR replaces the beam-size-pruning using `std::nth_element` with histogram pruning. The `beamSizePruning` and `scorePruning` functions of `LexiconfreeTimesyncBeamSearch`, `LexiconfreeLabelsyncBeamSearch` and `TreeTimesyncBeamSearch` have been combined into one function that can do both at once by incorporating the histogram.

The advantage in terms of speed is only minor; I measured around 1-2% search RTF reduction.

### DIFF
--- a/src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
+++ b/src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
@@ -125,6 +125,12 @@ const Core::ParameterFloat TreeTimesyncBeamSearch::paramWordEndScoreThreshold(
         If not set, global score pruning will be done and word-end hypotheses will not be pruned separately.",
         Core::Type<Score>::max, 0);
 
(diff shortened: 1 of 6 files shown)


src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
```diff
@@ -607,7 +609,7 @@ bool TreeTimesyncBeamSearch::decodeStep() {
     /*
      * Prune set of word-end extensions by max beam size and possibly also by score.
```
hannah220:
```suggestion
     * Prune set of word-end extensions by score.
```
max beam size is not enforced in the pruning below.

hannah220 (changes requested):

hannah220 (approved):

larissakl (approved):

PR #199, opened 2026-03-19
Add on-the-fly pre-pruning before creating extensions in SearchV2 algorithms

SimBe195:
This PR reduces the number of extensions that get created in the new search algorithms by skipping over options that are already worse than the best observed score so far + the first score threshold, i.e., extensions that would get pruned anyway. When skipping over them, no `ExtensionCandidate` object gets created and pushed into the list. The effect is relatively minor, e.g. 0.017 to 0.016 search RTF for a 10k BPE transducer in lexiconfree timesync search.


src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -404,27 +383,70 @@ bool LexiconfreeTimesyncBeamSearch::decodeStep() {
...
+                             .score          = extScore,
+                             .timeframe      = std::max(hyp.trace->time, (*scoreAccessor)->getTime()),
+                             .transitionType = transitionType,
+                             .baseHypIndex   = hypIndex});
+                }
+            }
         }
+        else {
+            // Remove extensions that couldn't be scored
+            extensions_.erase(
```
curufinwe: See below

larissakl (commented):
My comments in the LexiconfreeTimesyncBeamSearch also apply to the TreeTimesyncBeamSearch.

curufinwe (approved):

larissakl (approved):

PR #200, opened 2026-03-24
Add Onnx::ChunkedOnnxEncoder

SimBe195:
This PR adds a new Encoder class `ChunkedOnnxEncoder` which inherits from `OnnxEncoder`. The `OnnxEncoder` class has been slightly refactored so that the session run logic can be re-used in the child class. The `ChunkedOnnxEncoder` has a few additional parameters that define the structure of the chunks: "chunk-size", "step-size", "left-padding" and "right-padding".

For example with chunk-size = 50, step-size = 25, left-padding = 10, right-padding = 5 and 90 feature inputs in total
 - The first chunk consists of features 0 to 55 and the outputs corresponding to features 0 to 50 are returned
 - The second chunk consists of features 15 to 80 and the outputs corresponding to features 25 to 75 are returned
 - The third chunk consists of features 40 to 90 and the outputs corresponding to features 50 to 90 are returned

### DIFF
--- a/src/Onnx/OnnxEncoder.cc
+++ b/src/Onnx/OnnxEncoder.cc
@@ -192,6 +192,8 @@ const Core::Choice ChunkedOnnxEncoder::interpolationModeChoice(
         "no-interpolation", InterpolationMode::NoInterpolation,
         "linear", InterpolationMode::Linear,
         "linear-renorm", InterpolationMode::LinearRenorm,
+        "log-linear", InterpolationMode::LogLinear,
+        "log-linear-renorm", InterpolationMode::LogLinearRenorm,
         "neglog-linear", InterpolationMode::NegLogLinear,
         "neglog-linear-renorm", InterpolationMode::NegLogLinearRenorm,
         Core::Choice::endMark());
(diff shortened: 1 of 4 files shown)


src/Onnx/OnnxEncoder.cc
```diff
@@ -65,62 +71,182 @@ void OnnxEncoder::reset() {
...
-    for (size_t t = 0ul; t < T_in; ++t) {
-        std::copy(inputBuffer_[t].data(), inputBuffer_[t].data() + F, value.data<f32>(0, t));
+    for (size_t t = 0ul; t < nOutputs; ++t) {
+        size_t endInput = std::min(startInput + inputsPerOutput, nInputs);
+        outputBuffer_.push_back(
+                Nn::EncodedSpan{
+                        .encoding    = {outputView, outputSize, t * outputSize},
+                        .input_start = startInput,
+                        .input_end   = endInput});
+        startInput = std::min(startInput + inputStep, std::max(nInputs, 1ul) - 1ul);
```
hannah220, line 144: Should we rather break from for loop if `startInput + inputStep` is larther than `nInputs`?

src/Onnx/OnnxEncoder.cc
```diff
@@ -65,62 +71,182 @@ void OnnxEncoder::reset() {
...
+    Precursor::reset();
+    chunkCenterStart_     = 0ul;
+    numDiscardedFeatures_ = 0ul;
+}
+
+bool ChunkedOnnxEncoder::canEncode() const {
+    size_t availableEnd = numDiscardedFeatures_ + inputBuffer_.size();
+
+    if (not expectMoreFeatures_) {
+        return availableEnd > chunkCenterStart_;
```
hannah220, line 237:
```suggestion
        return availableEnd > chunkCenterStart_ + chunkSize_ + rightPadding_;
```

src/Onnx/OnnxEncoder.cc
SimBe195: Breaking the loop would skip outputs which is not desirable. I would rather have incorrect timestamps than skipping outputs. But this part anyway belongs to the original `OnnxEncoder` and not the new `ChunkedOnnxEncoder`.

src/Onnx/OnnxEncoder.cc
SimBe195: If no more features are coming, the last chunk doesn't need to be fully filled. We just need to have a non-zero amount of features in the chunk center to encode.

curufinwe (approved):
I only superficially looked through the code. But I did test it and in terms of output the new encoder matches the old behavior, so it's OK from my side.

hannah220 (approved):

PR #201, opened 2026-03-31
TreeTimesyncBeamSearch and LexiconfreeTimesyncBeamSearch: Fix timestamp increment at sentence end

hannah220:
Sentence End (SE) should have length 0, i.e., timeframe should not be consumed. 
However, since the label hypothesis constructor automatically increments timeframe, at `finalizeHypotheses`, SE hypotheses were created with timeframe which is out of range (because at this point all timeframes are consumed).
This PR simply decrement the timeframe in `finalizeHypotheses`.

### DIFF
--- a/src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
+++ b/src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
@@ -59,10 +59,12 @@ LexiconfreeTimesyncBeamSearch::LabelHypothesis::LabelHypothesis(
             predecessor = base.trace;
             break;
     }
-    trace = Core::ref(new LatticeTrace(
+    // Only increment timeframe when not SENTENCE_END
+    auto timeframe = Nn::TransitionType::SENTENCE_END ? extension.timeframe : extension.timeframe + 1;
+    trace          = Core::ref(new LatticeTrace(
             predecessor,
             extension.pron,
-            extension.timeframe + 1,
+            timeframe,
             {score, 0},
             {}));
 }
@@ -773,8 +775,6 @@ void LexiconfreeTimesyncBeamSearch::finalizeHypotheses() {
         auto&       ext     = extensions_[extensionIdx];
         auto const& baseHyp = beam_[ext.baseHypIndex];
         // The scoring context is not updated as no further scoring is done afterwards
-        // Make sentence-end length 0 as it should not consume a timestep
-        ext.timeframe -= 1;
         tempHypotheses_.push_back({baseHyp, ext, baseHyp.scoringContexts});
     }
 
--- a/src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
+++ b/src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
@@ -773,6 +773,8 @@ void LexiconfreeTimesyncBeamSearch::finalizeHypotheses() {
         auto&       ext     = extensions_[extensionIdx];
         auto const& baseHyp = beam_[ext.baseHypIndex];
         // The scoring context is not updated as no further scoring is done afterwards
+        // Make sentence-end length 0 as it should not consume a timestep
+        ext.timeframe -= 1;
         tempHypotheses_.push_back({baseHyp, ext, baseHyp.scoringContexts});
     }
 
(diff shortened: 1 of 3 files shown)


SimBe195:
Rather than decrementing the timestep later I would avoid incrementing it in the first place for sentence-end.

hannah220:
I see, so like this?
```
        extensions_.push_back(
                {.nextToken      = sentenceEndLabelIndex_,
                 .pron           = sentenceEndLemma_->pronunciations().first,
                 .score          = hyp.score,
                 .timeframe      = hyp.trace->time - 1,
                 .transitionType = Nn::TransitionType::SENTENCE_END,
                 .baseHypIndex   = hypIndex});
    }
```

larissakl:
I had the same idea as Simon. I think removing the line `ext.timeframe = std::max(ext.timeframe, (*scoreAccessors[extensionIdx])->getTime());` should do it.

SimBe195:
> I see, so like this?

I was thinking to handle sentence-end as a special case in the constructor but that also works.

hannah220:
I updated the label hyp constructor. 
In case of `TreeTimesyncBeamSearch`, the timeframe of LabelHypothesis will be the same as its trace. 
This is different to how previously done (only trace timeframe incremented) and I thought it is a bug, but let me know if it's wrong.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -59,10 +59,12 @@ LexiconfreeTimesyncBeamSearch::LabelHypothesis::LabelHypothesis(
             predecessor = base.trace;
             break;
     }
-    trace = Core::ref(new LatticeTrace(
+    // Only increment timeframe when not SENTENCE_END
+    auto timeframe = Nn::TransitionType::SENTENCE_END ? extension.timeframe : extension.timeframe + 1;
```
hannah220: This is anyway very wrong.. I'm not comparing `Nn::TransitionType::SENTENCE_END` with anything. 
I'll fix this.

hannah220:
A change to `WordEndExtensionCandidate` was neccesarry to include transition type.

hannah220:
I'm currently testing this.

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
```diff
@@ -66,17 +66,19 @@ TreeTimesyncBeamSearch::LabelHypothesis::LabelHypothesis(
...
           currentState(extension.rootState),
           lmHistory(newLmHistory),
-          timeframe(base.timeframe),
           score(extension.score) {
     auto newLmScore   = score - base.score;
     auto totalLmScore = base.trace->score.lm + newLmScore;
     auto totalAmScore = score - totalLmScore;
 
+    // Only increment timeframe when not SENTENCE_END
+    timeframe = extension.transitionType == Nn::TransitionType::SENTENCE_END ? base.timeframe : base.timeframe + 1;
```
larissakl: In case we don't have sentence end, you set `timeframe = base.timeframe + 1`. This means that you also increment the timeframe of the LabelHypothesis and not only the timeframe we use for the trace. So in the end the trace has the correct timeframe, but the hypothesis's timeframe is one too high. Is this possible or am I wrong about that?

larissakl: This only applies to the treesearch btw, the lexiconfree search looks fine.

hannah220: Shouldn't LabelHypothesis's timeframe actually also be incremented?
I tested both: as in this PR, and as you mentioned (do not change the LabelHypothesis's timeframe and only of the trace). 
The WERs of both cases in `ES_MBW_TEST_SET_SPECS_V3` were the same. 
So I guess both work but which one is algorithmically correct?

larissakl: Ah, I think I know why it does not make a difference. I initially thought that this would influence the scoring context of the hypothesis (such that the hyp is in a different timeframe and therefore gets scores from this wrong timeframe), but I forgot that the scoring context is not directly related to the timeframe field of LabelHypothesis, so the scoring is not influenced by that. The extensions' timeframes are also correct because of `ext.timeframe = std::max(ext.timeframe, scoreAccessor->getTime());`, which is why there are no subsequent problems. In my opinion, it is more correct to not increment the timeframe of a LabelHypothesis at a word end, but you are also right that it does not matter directly.

hannah220: Ok, thanks for looking into this! 
Then I'll not increment the timeframe of a LabelHypothesis at a word end.

larissakl (approved):

SimBe195 (approved):

PR #204, opened 2026-04-22
Refactoring: Insert explicit casts in many places to get rid of sign-compare warnings

larissakl:
My dream is to have RASR compiling without warnings. Or let's say, realistically, with only very few warnings.
Here is a first step. I went through all `Wsign-compare` warnings and inserted explicit casting to fix them.

However I can't guarantee that this is 100% safe. The problem is that if a signed value is casted to an unsigned value, but the value is negative, it will lead to an underflow. I think this should actually be fine everywhere, but I would appreciate if someone could double-check.

### DIFF
--- a/src/Onnx/Value.cc
+++ b/src/Onnx/Value.cc
@@ -70,7 +70,7 @@ void dynamic_rank_concat(Ort::Value& out, std::vector<Ort::Value const*> const& 
         data.push_back(values[value_idx]->GetTensorData<T>());
     }
 
-    for (size_t block_idx = 0ul; block_idx < static_cast<size_t>(num_blocks); block_idx++) {
+    for (int64_t block_idx = 0ul; block_idx < num_blocks; block_idx++) {
         int64_t partial_sum = 0l;
         for (size_t value_idx = 0ul; value_idx < data.size(); ++value_idx) {
             std::copy(data[value_idx] + block_sizes[value_idx] * block_idx, data[value_idx] + block_sizes[value_idx] * (block_idx + 1), data_out + (out_block_size)*block_idx + partial_sum);
--- a/src/Onnx/Value.cc
+++ b/src/Onnx/Value.cc
@@ -70,7 +70,7 @@ void dynamic_rank_concat(Ort::Value& out, std::vector<Ort::Value const*> const& 
         data.push_back(values[value_idx]->GetTensorData<T>());
     }
 
-    for (size_t block_idx = 0ul; block_idx < num_blocks; block_idx++) {
+    for (size_t block_idx = 0ul; block_idx < static_cast<size_t>(num_blocks); block_idx++) {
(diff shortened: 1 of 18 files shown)


SimBe195 (approved):

src/Onnx/Value.cc
```diff
@@ -70,7 +70,7 @@ void dynamic_rank_concat(Ort::Value& out, std::vector<Ort::Value const*> const&
         data.push_back(values[value_idx]->GetTensorData<T>());
     }
 
-    for (size_t block_idx = 0ul; block_idx < num_blocks; block_idx++) {
```
curufinwe, line 73: Maybe it's better to use int64_t for the iterator here, since the sizes are passed as int64_t?

src/Search/LanguageModelLookahead.cc
```diff
@@ -732,7 +732,7 @@ void LanguageModelLookahead::ConstructionTree::build(HMMStateNetwork const&
         s32 collectTopologicalStates(StateId node, int depth,
                                      std::vector<std::vector<StateId>>& topologicalStates,
                                      std::vector<s32>&                  collected) {
-            if (topologicalStates.size() <= depth) {
+            if (topologicalStates.size() <= static_cast<size_t>(depth)) {
```
curufinwe: Maybe we can instead just make depth an unsigned int? Depends how it is used.

src/Search/LanguageModelLookahead.cc
```diff
@@ -1291,7 +1291,7 @@ void LanguageModelLookahead::buildCompressesLookaheadStructure(u32
     log() << "lexicon syntactic tokens: " << lm_->lexicon()->nSyntacticTokens();
 
     // using lexicon_->nSyntacticTokens() intead of lm_->tokenInventory().size() here for mismatch case
-    for (int token = 0; token < lm_->lexicon()->nSyntacticTokens(); ++token) {
+    for (int token = 0; token < static_cast<int>(lm_->lexicon()->nSyntacticTokens()); ++token) {
```
curufinwe, line 1294: Can we maybe change the loop variable here? What does equal_range do?

larissakl: I don't think so. The key type of `nodeForTokenMap` is `Bliss::Token::Id`, so an implicit conversion at `equal_range(token)` would be introduced.

src/Search/LanguageModelLookahead.cc
larissakl: In general yes, but then we would also need to change the type of `collected`to `std::vector<size_t>&` and this in turn introduced `Wsign-compare` warning for other comparisons, for example in `verify(collected[node] != -2);`. So not sure if this really makes things easier.

curufinwe (approved):

PR #205, opened 2026-04-23
Add CtcPrefixLabelScorer

SimBe195:
This PR adds a new LabelScorer class for CTC prefix score computation. The way the scores are computed can also be found in Algorithm 2 "CTC label sequence score" of [Hybrid CTC/Attention Architecture for End-to-End Speech Recognition (Watanabe et al., 2017)](https://www.merl.com/publications/docs/TR2017-190.pdf).

Since this LabelScorer works label-synchronously, its main use is for combining CTC with other label-synchronous scorers such as AED in search.


src/Nn/LabelScorer/CtcPrefixLabelScorer.cc
```diff
@@ -0,0 +1,245 @@
...
+    auto ctcScorerContext = ctcScorer_->getInitialScoringContext();
+    while (true) {
+        auto scoreAccessor = ctcScorer_->getScoreAccessor(ctcScorerContext);
+
+        if (not scoreAccessor) {
+            break;
+        }
+
+        // Add column for next timestep to matrix and insert score values into it
+        ctcScores_->resizeColsAndKeepContent(ctcScores_->nColumns() + 1);
```
curufinwe: Wouldn't it be more efficient to iterate through the score accessors first to get the size and then create the full matrix in one go?

src/Nn/LabelScorer/CtcPrefixLabelScorer.cc
```diff
@@ -0,0 +1,245 @@
...
+
+ScoringContextRef CtcPrefixLabelScorer::extendedScoringContext(ScoringContextRef scoringContext, LabelIndex nextToken, TransitionType transitionType) {
+    auto context = Core::ref(dynamic_cast<CtcPrefixScoringContext const*>(scoringContext.get()));
+
+    std::vector<LabelIndex> newLabelSeq(context->labelSeq);
+    newLabelSeq.push_back(nextToken);
+
+    // Usually before extending the scores have already been computed in search, so this should just result in a cache lookup
+    auto scoreAccessor = getScoreAccessor(scoringContext);
+    verify(scoreAccessor);
```
curufinwe: This can fail as getScoreAccessor returns {} if expectMoreFeatures_ is true. Shouldn't we have better handling here?

src/Nn/LabelScorer/CtcPrefixLabelScorer.cc
```diff
@@ -0,0 +1,245 @@
...
+
+const Core::ParameterInt CtcPrefixLabelScorer::paramBlankIndex("blank-label-index", "Index of blank symbol in vocabulary.");
+const Core::ParameterInt CtcPrefixLabelScorer::paramVocabSize("vocab-size", "Number of labels in CTC scorer vocabulary.");
+
+CtcPrefixLabelScorer::CtcPrefixLabelScorer(Core::Configuration const& config)
+        : Core::Component(config),
+          Precursor(config, TransitionPresetType::LM),
+          blankIndex_(paramBlankIndex(config)),
+          vocabSize_(paramVocabSize(config)),
+          ctcScorer_(Module::instance().labelScorerFactory().createLabelScorer(select("ctc-scorer"))),
```
larissakl, line 135: Maybe we should somehow explicitly mention (like adding a comment in the header) that the "actual" scorer has to be set via "ctc-scorer" as this is not necessarily intuitive.

src/Nn/LabelScorer/CtcPrefixLabelScorer.cc
larissakl: However, if this happens, there is indeed something wrong, `expectMoreFeatures_` should not be true when `extendedScoringContext()` is called.

SimBe195: I agree that this is kind of ugly. We rely on the fact that in practice `getScoreAccessor` is always called first and `extendedScoringContext` later for surviving extensions. If someone calls the functions in a "wrong" order, it can crash. But there is not really a fail-mechanism for `extendedScoringContext`. It can't indicate success. We can either change the return-type, or change the `CtcPrefixScoringContext` to possibly delay setting `extScore` or just accept that this issue exists but doesn't come up in practice.

larissakl (approved):

src/Nn/LabelScorer/CtcPrefixLabelScorer.cc
curufinwe: I think the problem here is that we try to call getScoreAccessor here in the first place. We should simply move the score computation into the getScoreAccessor of this class. There we can simply return nullopt if we are not ready yet.

SimBe195: The reason it works like this is because the search expects the LabelScorer to return only the score for the next token. But the prefix-score that the algorithm computes is a sequence score. So for the score-delta it returns basically `score(prefix + next_token) - score(prefix)`. But to do that the `score(prefix)` value needs to be computed and stored first.

curufinwe (approved):

PR #209, opened 2026-04-24
Add functionality for common-prefix and n-best output to librasr SearchAlgorithm

SimBe195:
This PR extends the functionality of the librasr SearchAlgorithm pybinds. It adds 3 new functions:

1. `get_common_prefix` which returns the stable traceback and is useful for streaming applications
2. `get_current_n_best_list` which returns an n-best list instead of the single-best traceback
3. `recognize_segment_n_best` which is a convenience function like `recognize_segment` but returns the n-best instead of single-best.

To make the n-best retrieval easier, the conversion function from search-lattice to flf-lattice has been refactored out of the `Flf::RecognizerV2` class to make it reusable.


src/Flf/RecognizerV2.cc
```diff
@@ -104,23 +110,19 @@ void RecognizerNodeV2::work() {
     clog() << Core::XmlClose("layer");
 }
 
-ConstLatticeRef RecognizerNodeV2::buildLattice(Core::Ref<const Search::LatticeAdaptor> latticeAdaptor, std::string segmentName) {
-    auto lmScale = modelCombination_->languageModel()->scale();
-
+ConstLatticeRef convertSearchLatticeToFlf(Core::Ref<const Search::LatticeAdaptor> latticeAdaptor, Flf::LatticeHandler const* handler, std::string segmentName, f32 lmScale) {
```
curufinwe: Let's pass the lexicon in here instead of using the global one.

curufinwe (changes requested):

larissakl (approved):

curufinwe (approved):

PR #213, opened 2026-04-30
Add Module Torch

larissakl:


src/Core/Assertions.hh
```diff
@@ -168,8 +168,19 @@ template<class S, class T>
 void assertionFailedVerbose(const S& x, const T& y, const char* op,
                             const char* expr, const char* function,
                             const char* filename, unsigned int line) {
-    FailedAssertion("assertion", expr, function, filename, line).stream()
-            << x << " " << op << " " << y;
+    //FailedAssertion("assertion", expr, function, filename, line).stream()
+    //        << x << " " << op << " " << y;
+
```
curufinwe:
```suggestion
```

src/Core/Assertions.hh
```diff
@@ -168,8 +168,19 @@ template<class S, class T>
...
+    if constexpr (std::is_same_v<T, std::nullptr_t>) {
+        FailedAssertion("assertion", expr, function, filename, line).stream()
+                << x << " " << op << " " << "null";
+    }
+    else {
+        FailedAssertion("assertion", expr, function, filename, line).stream()
+                << x << " " << op << " " << y;
+    }
+
+    std::abort();
```
larissakl: Any suggestions for that?
I'm getting the compile error `error: ambiguous overload for ‘operator<<’ (operand types are ‘std::basic_ostream<char>’ and ‘std::nullptr_t’)` for everything which is like `require(x, nullptr)` (for example in src/Tools/Lm/LmUtilityTool.cc, line 161). This is currently my workaround, but maybe there is a better solution.

src/Nn/LabelScorer/DataView.hh
```diff
@@ -30,6 +30,13 @@
 #pragma pop_macro("ensure")
 #endif
 
+#ifdef MODULE_TORCH
+#pragma push_macro("ensure")
+#undef ensure
+#include <Torch/Tensor.hh>
+#pragma pop_macro("ensure")
+#endif
```
curufinwe, line 38: This macro thing was a TF specific workaround. Did you just copy & paste it or is it also needed for Torch?

src/Core/Assertions.hh
curufinwe: I think the solution is OK. My LLM friend suggests to remove the std::abort() and the commented out code. Another fix would be to change `require_ne(..., nullptr)` to `require_notnull(...)`. I think we can do both. The latter in a separate branch / PR.

cmake_resources/Torch.cmake
```diff
@@ -0,0 +1,80 @@
...
+        )
+    endif ()
+endif ()
+
+message(STATUS "Torch library directory: ${_torch_lib_dir}")
+message(STATUS "Torch include dirs: ${TORCH_INCLUDE_DIRS}")
+message(STATUS "Torch libraries: ${TORCH_LIBRARIES}")
+message(STATUS "Torch CXX flags: ${TORCH_CXX_FLAGS}")
+
+add_library(RasrExternalTorch INTERFACE)
```
curufinwe: We could give the library a proper name. E.g. LibTorch instead of Rasr... which denotes an internal module.

src/Nn/LabelScorer/DataView.hh
larissakl: It was needed for Torch as well. I don't remember the exact problem, but there some collision with the ensure macro.

SimBe195: The problem used to be that both numpy headers and RASR define an `ensure` macro, so the RASR one has to be temporarily disabled when including numpy headers to avoid conflicts.

PR #216, opened 2026-05-08
Update documentation on build system

larissakl:
I'm aware that the information in the section "Python integration" is also partly outdated, especially the reference to RETURNN. Any suggestions how to update that?


doc/architecture.rst
```diff
@@ -4,50 +4,74 @@ Architecture
...
-    Options.make
-    Rules.make
-    config/*.make
+    cmake -S . -B build
+    cmake --build build
+    cmake --install build
 
-The low-level ("atomic") makefiles are:
+The first command configures the project and generates the build system in ``build/``. The second command compiles the entire project or the target if selected via ``--target <NAME>``.
+The third command installs the executables and libraries below ``arch/<system>-<architecture>-<build-type>/``, unless a different installation prefix is selected.
```
curufinwe, line 14: I don't think it's configurable atm. But we can change that.

curufinwe (changes requested):

doc/architecture.rst
curufinwe: I updated the PR s.t. the user can actually set the prefix.

larissakl: Thank you!

curufinwe (approved):

SimBe195 (approved):

PR #217, opened 2026-05-08
Update new search algorithms: Perform all reset logic in enterSegment instead of reset

SimBe195:
Previously, the `enterSegment` function in the new search algorithms only reset the LabelScorers but kept the current hypotheses in the beam. For certain LabelScorers this can become a problem because the existing scoring contexts in the hypotheses can become invalid when resetting the LabelScorer.

This PR calls the `reset` function in `enterSegment` which first resets the LabelScorers and then also re-initializes the beam.

### DIFF
--- a/src/Flf/RecognizerV2.cc
+++ b/src/Flf/RecognizerV2.cc
@@ -48,7 +48,6 @@ void RecognizerNodeV2::recognizeSegment(const Bliss::SpeechSegment* segment) {
     }
 
     // Initialize recognizer and feature extractor
-    searchAlgorithm_->reset();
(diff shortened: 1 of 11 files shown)


larissakl:
Should we then maybe remove the call to `reset()` in `setModelCombination()`? It's actually not necessary anymore.

SimBe195:
> Should we then maybe remove the call to `reset()` in `setModelCombination()`? It's actually not necessary anymore.

It shouldn't be necessary but it also doesn't really hurt as a failsafe. I would be fine with both options.

larissakl (approved):

curufinwe (approved):

PR #206, opened 2026-04-23
Add StateMangedOnnxLabelScorer

SimBe195:
This PR adds a new LabelScorer class which utilizes a RASR state manager. In many ways this new scorer is similar to the StatefulOnnxLabelScorer but instead of the state-initializer/state-updater/scorer logic which is agnostic to the shapes of the states, this one uses the pattern from the RecurrentLanguageModel in RASR and has a single ONNX module plus a state manager.

The state manager and compressed vector factory creation have been moved from Lm to Nn in order to make them reusable for this LabelScorer without extra dependencies.

PR #218, opened 2026-05-08
Add PrefixSpeechLmOnnxLabelScorer

SimBe195:
This PR adds a new LabelScorer which can be used for Speech LLMs. Structurally it is roughly similar to the `StateManagedOnnxLabelScorer` from #206. The main difference is how the initial states are constructed. While the `StateMangedOnnxLabelScorer` only gets one or more start labels as initial history while the `PrefixSpeechLmOnnxLabelScorer` has a separate initializer ONNX model which gets an `initial-prompt`, `encoder-states` and `suffix-prompt` to produce the initial scores and states. Also, the `StateManagedOnnxLabelScorer` feeds in the encoder-states at every scoring step while the `PrefixSpeechLmOnnxLabelScorer` only feeds the encoder-states for the initialization.


src/Nn/LabelScorer/ScoringContext.hh
```diff
@@ -150,41 +148,6 @@ struct OnnxHiddenStateScoringContext : public ScoringContext {
...
-    CtcPrefixScoringContext(std::vector<LabelIndex>&& seq, ScoringContextRef const& parent)
-            : labelSeq(std::move(seq)), parent(parent), timePrefixScores(), prefixScore(), extScores(), requiresFinalize(true) {}
-
-    bool   isEqual(ScoringContextRef const& other) const;
-    size_t hash() const;
-};
-
-typedef Core::Ref<const CtcPrefixScoringContext> CtcPrefixScoringContextRef;
-
 /*
```
hannah220: According to https://github.com/rwth-i6/rasr/pull/229, let's move these to respective label scorer headers.

src/Nn/LabelScorer/StateManagedOnnxLabelScorer.hh
```diff
@@ -25,7 +25,7 @@
 namespace Nn {
 
 /*
- * LabelScorer for ONNX models whose hidden-state management is done by a RASR StateManager.
+ * LabelScorer for ONNX models whose hidden-state management done by a Rasr StateManager.
```
hannah220:
```suggestion
 * LabelScorer for ONNX models whose hidden-state management is done by a Rasr StateManager.
```

src/Nn/LabelScorer/StateManagedOnnxLabelScorer.cc
hannah220, line 395: Why from `outputs.begin() + 1` (the first element will be igonred)?

src/Nn/LabelScorer/PrefixSpeechLmOnnxLabelScorer.cc
```diff
@@ -0,0 +1,494 @@
...
+                        initialPromptLabels_.data(),
+                        std::vector<s64>{1l, static_cast<s64>(initialPromptLabels_.size())}));
+        if (initializerInitialPromptLengthName_ != "") {
+            inputs.emplace_back(
+                    initializerInitialPromptLengthName_,
+                    Onnx::Value::create(
+                            std::vector<s32>{static_cast<s32>(initialPromptLabels_.size())}));
+        }
+    }
+
```
hannah220, line 394: Don't we need to check if `initializerEncoderStatesName_` is true before calling `setupEncoderStatesValue()`?

src/Nn/LabelScorer/PrefixSpeechLmOnnxLabelScorer.cc
```diff
@@ -0,0 +1,494 @@
...
+    outputs.front().get(0, *scores);
+
+    std::vector<Onnx::Value> stateOutputs(std::make_move_iterator(outputs.begin() + 1), std::make_move_iterator(outputs.end()));
+    std::vector<size_t>      suffixLengths{stateLength};
+    auto                     promptStates = stateManager_->splitStates(stateVariables_, suffixLengths, stateOutputs, *stateVectorFactory_);
+
+    PrefixSpeechLmScoringContextRef parent;
+    if (stateManager_->requiresAllParentStates()) {
+        verify_eq(promptStates.size(), stateLength);
+        for (size_t i = 0ul; i + 1ul < promptStates.size(); ++i) {
```
hannah220, line 448: Why not until `i < promptStates.size()`?

src/Nn/LabelScorer/StateManagedOnnxLabelScorer.cc
SimBe195: The first output is always the score vector and the ones after are states.

src/Nn/LabelScorer/ScoringContext.hh
SimBe195: Done.

src/Nn/LabelScorer/PrefixSpeechLmOnnxLabelScorer.cc
SimBe195: Currently the encoder-state input is not optional. It should in principle also be possible to make it optional here, but for text-only prompt without encoder states, one should typically use a `StateManagedOnnxLabelScorer` instead of this one.

src/Nn/LabelScorer/PrefixSpeechLmOnnxLabelScorer.cc
SimBe195: The last element of `promptStates` is assigned to `initialContext->state` further down. This loop only builds the states for the parents.

PR #220, opened 2026-05-11
Update Python.cmake: Initialize FindPython before pybind11 is used

larissakl:
Recent pybind11 CMake code uses CMake's modern `FindPython` support for `pybind11_add_module`. In a new Apptainer image I was using, configuration failed in `pybind11NewTools.cmake` with `Unknown CMake command "python_add_library"` because the generic Python package had not been initialized with module development support.

I only ran into this problem when having the new Torch module enabled, since the Torch setup also queries the active Python environment.

### DIFF
--- a/cmake_resources/Python.cmake
+++ b/cmake_resources/Python.cmake
@@ -1,5 +1,4 @@
-find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
-find_package(Python3 REQUIRED COMPONENTS Development NumPy)
+find_package(Python3 REQUIRED COMPONENTS Development NumPy Interpreter Development.Module)
 
 add_library(RasrPythonDependencies INTERFACE)
 target_compile_definitions(
--- a/cmake_resources/Python.cmake
+++ b/cmake_resources/Python.cmake
@@ -1,10 +1,9 @@
-find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
-find_package(Python3 REQUIRED COMPONENTS Development NumPy)
+find_package(Python REQUIRED COMPONENTS Development NumPy Interpreter Development.Module)
 
 add_library(RasrPythonDependencies INTERFACE)
 target_compile_definitions(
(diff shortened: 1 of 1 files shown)


cmake_resources/Python.cmake
```diff
@@ -1,3 +1,4 @@
+find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
 find_package(Python3 REQUIRED COMPONENTS Development NumPy)
```
SimBe195: If python3 exists on the system, `find_package(Python` is equivalent to `find_package(Python3`, so I would put all components under one `find_package` command.

SimBe195 (approved):

larissakl:
@SimBe195 I just tested it again (I haven't done it before committing this, my bad) and it only works if I have them separated as before, so
```
find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
find_package(Python3 REQUIRED COMPONENTS Development NumPy)
```
I don't know why though.

SimBe195:
> @SimBe195 I just tested it again (I haven't done it before committing this, my bad) and it only works if I have them separated as before, so
> I don't know why though.

What if you put both under `find_package(Python...)`?

larissakl:
> What if you put both under find_package(Python...)?

Then I'm getting the error 

```
CMake Error at cmake_resources/Python.cmake:8 (target_link_libraries):
 The link interface of target "RasrPythonDependencies" contains:

   Python3::Python

 but the target was not found.
```

SimBe195:
> > What if you put both under find_package(Python...)?
> 
> Then I'm getting the error
> ...


Ah, if you use `find_package(Python...)` the imported targets are also named `Python:...`, so change the links later to
```
target_link_libraries(
    RasrPythonDependencies INTERFACE Python::Python Python::NumPy
)
```

larissakl:
Works fine now, thank you :+1:

curufinwe (approved):

PR #223, opened 2026-05-15
Add dense score retrieval for certain ScoreAccessors

SimBe195:
Some ScoreAccessors (e.g. `VectorScoreAccessor`) represent a dense span of score values over the vocabulary that don't depend on the transition type. For these ScoreAccessor classes we can implement an optimized function to retrieve the full dense score span directly and thus fetch all scores at once. This avoids repeated virtual `scoreAccessor->getScore(transitionType, labelIndex)` function calls in the search algorithms for every possible extension. In the future it can also be used to introduce local pruning over the dense score span when it's available.

Originally I modeled the DenseScoreSpan just as a `std::span<Score const>` but that doesn't support scaling and interpolation as it is used in `CombineLabelScorer`, `PriorLabelScorer` and `ScaledLabelScorer`. So now it is instead modeled as a struct containing multiple terms each of which consist of a span + a scale.

In a test segment with a speech LLM and lexiconfree labelsync search, the dense score accessors reduce the search time from 13.39s to 12.85s (RTF 0.336 to 0.323).


src/Nn/LabelScorer/ScoreAccessor.hh
```diff
@@ -16,43 +16,76 @@
...
+    std::vector<DenseScoreTerm> terms;
+
+    DenseScoreSpan(std::vector<DenseScoreTerm>&& terms)
+            : terms(std::move(terms)) {}
+    DenseScoreSpan(DenseScoreTerm&& term)
+            : terms{std::move(term)} {};
+
+    size_t size() const {
+        return terms.empty() ? 0ul : terms.front().scores.size();
+    }
```
larissakl, line 56: I think this might be confusing because one would usually assume to get the size of the vector `terms` when calling `size()` for this struct, not size of the scores in `terms`. Maybe we could rename the function to have it more explicit?

src/Nn/LabelScorer/ScaledLabelScorer.cc
```diff
@@ -32,6 +32,19 @@ class ScaledScoreAccessor : public ScoreAccessor {
...
 
+    std::optional<DenseScoreSpan> getDenseScores() const override {
+        auto denseScores = base_->getDenseScores();
+        if (scale_ == 1.0 or not denseScores) {
+            return denseScores;
+        }
+
+        std::vector<DenseScoreTerm> denseScoreTerms(denseScores->terms.begin(), denseScores->terms.end());
+        for (auto& term : denseScoreTerms) {
+            term.scale *= scale_;
```
larissakl, line 42: Why `*=` and not `=`?

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -414,8 +424,13 @@ bool LexiconfreeTimesyncBeamSearch::decodeStep() {
...
                     auto extTime        = hyp.trace->time;
                     if (labelScorers_[scorerIdx]->scoresTransition(transitionType)) {
-                        extScore += (*scoreAccessor)->getScore(transitionType, tokenIdx);
-                        extTime = std::max(extTime, (*scoreAccessor)->getTime());
+                        if (denseScores and tokenIdx < denseScores->size()) {
+                            extScore += (*denseScores)[tokenIdx];
+                        }
+                        else {
+                            extScore += (*scoreAccessor)->getScore(transitionType, tokenIdx);
+                        }
```
larissakl: Why do you use an if-else block here and not the same syntax as below?

src/Nn/LabelScorer/CombineLabelScorer.cc
```diff
@@ -40,6 +40,30 @@ class CombinedScoreAccessor : public ScoreAccessor {
...
+            if (not denseScores) {
+                return std::nullopt;
+            }
+            if (denseScoreTerms.empty()) {
+                denseSize = denseScores->size();
+            }
+            else if (denseScores->size() != denseSize) {
+                return std::nullopt;
+            }
+            denseScoreTerms.insert(denseScoreTerms.end(), denseScores->terms.begin(), denseScores->terms.end());
```
larissakl, line 99: I don't get the implementation for this ScoreAccessor. Why do you just collect the dense score terms of all subAccessors in one vector? Don't we need to sum them up?

src/Nn/LabelScorer/ScoreAccessor.hh
curufinwe: I don't mind either way, but it is consistent in the view that it returns the maximum index + 1 for the array accessor below. In that sense it is like a vector.

src/Nn/LabelScorer/ScoreAccessor.hh
```diff
@@ -16,43 +16,76 @@
...
+};
+
+/*
+ * Dense score view over one or more contiguous vocabulary score spans.
+ */
+struct DenseScoreSpan {
+    std::vector<DenseScoreTerm> terms;
+
+    DenseScoreSpan(std::vector<DenseScoreTerm>&& terms)
+            : terms(std::move(terms)) {}
```
curufinwe: Maybe we could add an assertion here to check that all terms are of the same length.

src/Nn/LabelScorer/ScaledLabelScorer.cc
curufinwe: Because the terms might come from a combined label scorer that does scaling differently for each term. Then the scaled label scorer here should multiply its own scale and not override the scales of the combined label scorer.

src/Nn/LabelScorer/ScaledLabelScorer.cc
```diff
@@ -32,6 +32,19 @@ class ScaledScoreAccessor : public ScoreAccessor {
...
+        }
+        return DenseScoreSpan(std::move(denseScoreTerms));
```
curufinwe: Why do we need to create a new DenseScoreSpan here instead of returning the modified one?

src/Nn/LabelScorer/PriorLabelScorer.cc
```diff
@@ -26,14 +26,36 @@ class PriorScoreAccessor : public Nn::ScoreAccessor {
...
     }
 
-    virtual Nn::TimeframeIndex getTime() const {
+    std::optional<Nn::DenseScoreSpan> getDenseScores() const override {
+        auto denseScores = scoreAccessor_->getDenseScores();
+        if (not denseScores) {
+            return std::nullopt;
+        }
+
+	std::vector<Nn::DenseScoreTerm> denseScoreTerms(denseScores->terms.begin(), denseScores->terms.end());
```
curufinwe:
```suggestion
        std::vector<Nn::DenseScoreTerm> denseScoreTerms(denseScores->terms.begin(), denseScores->terms.end());
```

But actually: why do we need a copy? We could also just update denseScores, no?

src/Nn/LabelScorer/PriorLabelScorer.cc
```diff
@@ -26,14 +26,36 @@ class PriorScoreAccessor : public Nn::ScoreAccessor {
...
         return scoreAccessor_->getTime();
     }
 
 private:
-    Core::Ref<ScoreAccessor>          scoreAccessor_;
-    const bool                        negateInput_;
-    std::shared_ptr<Nn::Prior<Score>> prior_;
+    Core::Ref<ScoreAccessor>                scoreAccessor_;
+    const bool                              negateInput_;
+    std::shared_ptr<Nn::Prior<Score>>       prior_;
```
curufinwe: why this whitespace change?

src/Nn/LabelScorer/PriorLabelScorer.cc
```diff
@@ -26,14 +26,36 @@ class PriorScoreAccessor : public Nn::ScoreAccessor {
...
+                term.scale *= -1.0;
+            }
+        }
+
+        if (prior_->scale() != 0.0) {
+            require(denseScores->size() == prior_->size());  // Prior size must match base scorer vocab size
+            denseScoreTerms.push_back(Nn::DenseScoreTerm{denseScores->size() > 0ul ? std::span<Score const>(&prior_->at(0), prior_->size()) : std::span<Score const>(), prior_->scale()});
+        }
+
+        return Nn::DenseScoreSpan(std::move(denseScoreTerms));
```
curufinwe: Why create a new DenseScoreSpan instead of returning the modified one?

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
curufinwe: Shouldn't we also use the dense scores here?

curufinwe (changes requested):

src/Nn/LabelScorer/PriorLabelScorer.cc
```diff
@@ -26,14 +26,36 @@ class PriorScoreAccessor : public Nn::ScoreAccessor {
...
+
+        if (negateInput_) {
+            for (auto& term : denseScoreTerms) {
+                term.scale *= -1.0;
+            }
+        }
+
+        if (prior_->scale() != 0.0) {
+            require(denseScores->size() == prior_->size());  // Prior size must match base scorer vocab size
+            denseScoreTerms.push_back(Nn::DenseScoreTerm{denseScores->size() > 0ul ? std::span<Score const>(&prior_->at(0), prior_->size()) : std::span<Score const>(), prior_->scale()});
```
hannah220: In the former case, don't you also need to set the prior scale?
`std::span<Score const>(&prior_->at(0), prior_->size()), prior_->scale()`

src/Nn/LabelScorer/CombineLabelScorer.cc
```diff
@@ -40,6 +40,30 @@ class CombinedScoreAccessor : public ScoreAccessor {
...
+        size_t                      denseSize = 0ul;
+        for (auto const& subAccessor : subAccessors_) {
+            auto denseScores = subAccessor->getDenseScores();
+            if (not denseScores) {
+                return std::nullopt;
+            }
+            if (denseScoreTerms.empty()) {
+                denseSize = denseScores->size();
+            }
+            else if (denseScores->size() != denseSize) {
```
hannah220, line 96: denseSize will always be 0 in the `else if` above, right?

src/Nn/LabelScorer/CombineLabelScorer.cc
```diff
@@ -40,6 +40,30 @@ class CombinedScoreAccessor : public ScoreAccessor {
...
+        }
+
+        std::vector<DenseScoreTerm> denseScoreTerms;
+        size_t                      denseSize = 0ul;
+        for (auto const& subAccessor : subAccessors_) {
+            auto denseScores = subAccessor->getDenseScores();
+            if (not denseScores) {
+                return std::nullopt;
+            }
+            if (denseScoreTerms.empty()) {
```
hannah220, line 93: In which case `denseScoreTerms` is empty, but `denseScores->size()` will be some other size?

src/Nn/LabelScorer/ScoreAccessor.hh
SimBe195: From the outside, the number of terms doesn't really matter. This return basically the index until which `operator[]` can be called. I added a docstring to the methods to clarify.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
SimBe195: Unified this now.

src/Nn/LabelScorer/CombineLabelScorer.cc
SimBe195: Because the summation is done on-demand in `DenseScoreSpan::operator[]`.

src/Nn/LabelScorer/ScoreAccessor.hh
SimBe195: Done.

src/Nn/LabelScorer/ScaledLabelScorer.cc
SimBe195: That's right, copying here is unnecessary. Originally I think I didn't want to modify the base score accessor because calling `getDenseScores` multiple times would repeatedly apply the scale. But since `denseScores` is retrieved by value, this isn't a concern anymore.

src/Nn/LabelScorer/PriorLabelScorer.cc
SimBe195: Removed the copy.

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
SimBe195: Yes, did that now.

src/Nn/LabelScorer/CombineLabelScorer.cc
SimBe195: In the first loop iteration `denseScoreTerms` is empty and `denseSize` is 0. So the `if (denseScoreTerms.empty())` block triggers and `denseSize = denseScores->size()` is set. In every future loop iteration, the `else` path is taken and `denseSize != 0`.

src/Nn/LabelScorer/PriorLabelScorer.cc
SimBe195: I believe the prior scale is applied correctly. I changed the code a bit now to make it easier to read.

curufinwe (approved):

hannah220 (approved):

PR #229, opened 2026-06-01
Move specific ScoringContext classes to corresponding LabelScorers

SimBe195:
Many ScoringContext classes are only used by a single LabelScorer. This PR moves these from `ScoringContexts.[hh|cc]` next to the corresponding LabelScorer classes that use them. In addition, since many ScoringContexts have hashes and equality operators based on label sequences, helper functions `labelSeqHash` and `labelSeqEqual` are introduced in order to avoid repeated logic.

### DIFF
--- a/src/Nn/LabelScorer/FixedContextOnnxLabelScorer.cc
+++ b/src/Nn/LabelScorer/FixedContextOnnxLabelScorer.cc
@@ -18,6 +18,34 @@
 
 namespace Nn {
 
+SeqStepScoringContext::SeqStepScoringContext()
+        : labelSeq(), currentStep(0ul) {}
+
+SeqStepScoringContext::SeqStepScoringContext(std::vector<LabelIndex> const& seq, Speech::TimeframeIndex step)
+        : labelSeq(seq), currentStep(step) {}
+
+SeqStepScoringContext::SeqStepScoringContext(std::vector<LabelIndex>&& seq, Speech::TimeframeIndex step)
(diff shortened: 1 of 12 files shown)


src/Nn/LabelScorer/FixedContextOnnxLabelScorer.cc
```diff
@@ -18,6 +18,34 @@
...
+    }
+
+    if (currentStep != otherPtr->currentStep) {
+        return false;
+    }
+
+    return labelSeqEqual(labelSeq, otherPtr->labelSeq);
+}
+
+size_t SeqStepScoringContext::hash() const {
```
hannah220, line 43: I'm wondering that it's _equal_ if label seqs are the same, but the hash could be different as it also depends on the currentStep.

curufinwe: Good catch!

curufinwe (changes requested):

src/Nn/LabelScorer/FixedContextOnnxLabelScorer.cc
larissakl: Maybe I misunderstand something, but in line 36 we also check for currentStep, so both currentStep and the label seq must be equal.

curufinwe (approved):

larissakl (approved):

PR #228, opened 2026-05-29
Add silence TransitionTypes and HMM TransitionPreset

larissakl:
Adds new transitions types with silence (label-to-silence, silence-to-label, silence-loop and initial-silence) and a transition preset for HMM. Additionally, I added a `toString()` function for the TransitionTypes to make debugging easier.

Label scorers with a `blank-updates-history` parameter got a parallel `silence-updates-history` parameter.

I will refactor the `inferTransitionType()` functions in a separate PR.

PR #230, opened 2026-06-02
Refactor inferTransitionType: Use new silence transition types and take state change in search tree into account

larissakl:
Refactoring of the `inferTransitionType()` functions in the timesync searches.
The new silence transition types from #228 are now used and in the tree search a label-loop transition can only occur if the state in the search tree did not change.


src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
```diff
@@ -232,6 +244,18 @@ bool LexiconfreeTimesyncBeamSearch::setModelCombination(Speech::ModelCombination
...
     }
 
+    auto silenceLemma = lexicon_->specialLemma("silence");
+    if (silenceLemma) {
+        if (silenceLabelIndex_ == Nn::invalidLabelIndex) {
+            silenceLabelIndex_ = silenceLemma->id();
+            useSilence_        = true;
+            log() << "Use silence index " << silenceLabelIndex_ << " inferred from lexicon";
+        }
+        else if (silenceLabelIndex_ != static_cast<Nn::LabelIndex>(silenceLemma->id())) {
```
hannah220, line 254: It's also the case for blank, but if we say that it's overwritten by `silenceLabelIndex_`, `useSilence_` should be set to true here too because we intend to use that index as silence?

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
```diff
@@ -773,14 +785,32 @@ void TreeTimesyncBeamSearch::logStatistics() const {
...
+    bool nextIsSilence = (useSilence_ and nextLabel == silenceLabelIndex_);
+
+    if (not stateChanged) {
+        if (prevIsBlank) {
+            return Nn::TransitionType::BLANK_LOOP;
+        }
+        if (prevIsSilence) {
+            return Nn::TransitionType::SILENCE_LOOP;
+        }
+        else if (collapseRepeatedLabels_) {
```
hannah220, line 803:
```suggestion
        if (collapseRepeatedLabels_) {
```
Because if above returns.

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
```diff
@@ -773,14 +785,32 @@ void TreeTimesyncBeamSearch::logStatistics() const {
     numActiveTrees_.write(clog());
 }
 
-Nn::TransitionType TreeTimesyncBeamSearch::inferTransitionType(Nn::LabelIndex prevLabel, Nn::LabelIndex nextLabel) const {
+Nn::TransitionType TreeTimesyncBeamSearch::inferTransitionType(Nn::LabelIndex prevLabel, Nn::LabelIndex nextLabel, bool stateChanged) const {
```
hannah220: Minor suggestion: reverse bool `stateChanged` -> `stateStayed`/`sameState` or so,
and at line 795 just: `if (stateStayed)`

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
hannah220, line 819: Below there are a bit of duplicates, BLANK_LOOP and SILENCE_LOOP would be already returned above. So it could be cleaned up.

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
```diff
@@ -773,14 +785,32 @@ void TreeTimesyncBeamSearch::logStatistics() const {
...
     bool prevIsBlank = (useBlank_ and prevLabel == blankLabelIndex_);
     bool nextIsBlank = (useBlank_ and nextLabel == blankLabelIndex_);
 
+    bool prevIsSilence = (useSilence_ and prevLabel == silenceLabelIndex_);
+    bool nextIsSilence = (useSilence_ and nextLabel == silenceLabelIndex_);
+
```
hannah220, line 795: This means that silences and blanks are always collapsed regardless of `collapseRepeatedLabels_`?

larissakl: Yes, but this was also the case before.

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
larissakl: My idea here is that BLANK_LOOP and SILENCE_LOOP should be independent from the state change, so no matter if the state in the tree changed or not, we will have a BLANK_LOOP if we predict two consecutive blanks. However, in reality, it should usually never happen that we have two consecutive blanks with a state change.

src/Search/LexiconfreeTimesyncBeamSearch/LexiconfreeTimesyncBeamSearch.cc
larissakl: But I think in this case `useSilence_` is already set to true, isn't it? cf. line 201

hannah220: Oh you're right!

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
```diff
@@ -479,14 +490,15 @@ bool TreeTimesyncBeamSearch::decodeStep() {
                 for (size_t i = stateSuccessorsOffset_[hyp.currentState]; i < stateSuccessorsOffset_[hyp.currentState + 1]; ++i) {
                     const StateId  successorState = stateSuccessors_[i];
                     Nn::LabelIndex tokenIdx       = network_->structure.state(successorState).stateDesc.acousticModel;
-                    // If we collapse repeated labels, a new word should not start with the same token as the previous word ended (except for blank itself)
+                    // If we collapse repeated labels, a new word should not start with the same token as the previous word ended (except for blank or silence)
                     if (collapseRepeatedLabels_ and
                         hyp.currentState == network_->rootState and
```
curufinwe: network_->rootState might not be the only root state. I suggest you use the isRoot function to check for this: https://github.com/rwth-i6/rasr/blob/master/src/Search/PersistentStateTree.hh#L73

curufinwe (changes requested):

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
larissakl: Good point. However this check doesn't take `otherRootStates` into account. Should we maybe add this? Then we could maybe also adjust this check `if (hyp.currentState == network_->rootState or network_->otherRootStates.find(hyp.currentState) != network_->otherRootStates.end())` in` recombination()`

curufinwe: OK, let's extend this check and use it in recombination too.

src/Search/TreeTimesyncBeamSearch/TreeTimesyncBeamSearch.cc
larissakl: I think in this case it doesn't make a difference, does it?

curufinwe (approved):

hannah220 (approved):

PR #236, opened 2026-06-09
Add Python bindings to manipulate the LM/LabelScorer scale in SearchV2

larissakl:
Adds two new Python bindings for SearchAlgorithmV2 which act as setter and getter for the LM scale. In this way, the LM scale can be changed during the recognition without adjusting the config file, it just overwrites the value set in the config.

We need this for setups where the search is used during training to allow for a dynamic LM scale across the epochs. For standard recognition this might however not be very useful, so I would also be okay if you reject this change.

### DIFF
--- a/src/Nn/LabelScorer/CombineLabelScorer.cc
+++ b/src/Nn/LabelScorer/CombineLabelScorer.cc
@@ -69,6 +69,28 @@ CombineLabelScorer::CombineLabelScorer(Core::Configuration const& config, ModelC
     }
 }
 
+size_t CombineLabelScorer::numLabelScorers() const {
+    return scorers_.size();
+}
+
+Score CombineLabelScorer::scale(size_t index) const {
+    require(index < scorers_.size());
+
+    auto* scorer = dynamic_cast<ScaledLabelScorer*>(scorers_[index].get());
+    require(scorer);
+
+    return scorer->scale();
+}
+
+void CombineLabelScorer::setScale(size_t index, Score scale) {
+    require(index < scorers_.size());
+
+    auto* scorer = dynamic_cast<ScaledLabelScorer*>(scorers_[index].get());
+    require(scorer);
+
+    scorer->setScale(scale);
+}
+
 void CombineLabelScorer::reset() {
     for (auto& scorer : scorers_) {
         scorer->reset();
--- a/src/Nn/LabelScorer/CombineLabelScorer.cc
+++ b/src/Nn/LabelScorer/CombineLabelScorer.cc
@@ -108,6 +108,7 @@ CombineLabelScorer::CombineLabelScorer(Core::Configuration const& config, ModelC
 }
 
 Core::Ref<ScaledLabelScorer> CombineLabelScorer::getSubScorer(size_t index) const {
+    require(index < scorers_.size());
     return scorers_[index];
 }
 
--- a/src/Nn/LabelScorer/CombineLabelScorer.cc
+++ b/src/Nn/LabelScorer/CombineLabelScorer.cc
@@ -69,26 +69,8 @@ CombineLabelScorer::CombineLabelScorer(Core::Configuration const& config, ModelC
     }
 }
 
-size_t CombineLabelScorer::numLabelScorers() const {
-    return scorers_.size();
-}
-
-Score CombineLabelScorer::scale(size_t index) const {
-    require(index < scorers_.size());
-
-    auto* scorer = dynamic_cast<ScaledLabelScorer*>(scorers_[index].get());
-    require(scorer);
-
-    return scorer->scale();
-}
-
-void CombineLabelScorer::setScale(size_t index, Score scale) {
-    require(index < scorers_.size());
-
-    auto* scorer = dynamic_cast<ScaledLabelScorer*>(scorers_[index].get());
-    require(scorer);
-
-    scorer->setScale(scale);
+Core::Ref<ScaledLabelScorer> CombineLabelScorer::getSubScorer(size_t index) const {
+    return scorers_[index];
 }
 
 void CombineLabelScorer::reset() {
--- a/src/Nn/LabelScorer/CombineLabelScorer.hh
+++ b/src/Nn/LabelScorer/CombineLabelScorer.hh
@@ -18,6 +18,7 @@
 
 #include "LabelScorer.hh"
 #include "ModelCache.hh"
+#include "ScaledLabelScorer.hh"
 
 namespace Nn {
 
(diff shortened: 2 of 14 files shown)


curufinwe:
My suggestion would be to provide a getter for ModelCombination and then have the python code modify the language-model scale directly instead adding a pass-through function in Search.

larissakl:
I completely refactored my initial code and added pybinds for the ModelCombination, the LM, the ScaledLabelScorer and the CombineLabelScorer, such that the LM scale, the label scorer scale as well as the scales of the sub-scorers can be manipulated in Python.
@curufinwe Does this more or less fulfill your idea?

curufinwe (approved):

src/Nn/LabelScorer/CombineLabelScorer.hh
hannah220: Change LabelScorer to ScaledLabelScorer

larissakl:
I added functions to the CtcPrefixLabelScorer, the CombineLabelScorer and the EncoderDecoderLabelScorer to return their wrapped label scorers as ScaledLabelScorer. Now there are pybinds for the function getLabelScorer(modelCombination, index) which returns the modelCombination's label scorer at index and for the function getSubScorer(labelScorer, index) which returns the sub-scorer of the labelScorer (at index if it's a CombineLabelScorer). The label scorers are always returned as ScaledLabelScorers, to its functions scale() and getScale() can then be used directly.

To make this possible, I had to change the construction of the encoder-only label scorer. It used the EncoderDecoderLabelScorer with the decoder being a no-op label scorer. Then however, I couldn't return the decoder as ScaledLabelScorer. Therefore, the decoder's no-op label is now also wrapped by a ScaledLabelScorer. Theoretically it is therefore possible to set a scale for the decoder, but if nothing is set the behavior should be unchanged.

curufinwe (approved):

src/Python/LabelScorer.hh
```diff
@@ -19,10 +19,37 @@
...
+#include <Nn/LabelScorer/CtcPrefixLabelScorer.hh>
+#include <Nn/LabelScorer/EncoderDecoderLabelScorer.hh>
 #include <Nn/LabelScorer/LabelScorer.hh>
+#include <Nn/LabelScorer/ScaledLabelScorer.hh>
 
 namespace py = pybind11;
 
+// Return the sub-scorer that is wrapped by the given label scorer
+// If the label scorer is a CombineLabelScorer, return the sub-scorer at the specified index
+static Nn::ScaledLabelScorer* getSubScorer(Nn::ScaledLabelScorer* labelScorer, std::optional<size_t> index = 0) {
```
SimBe195: In my opinion this function doesn't really fit here. It would be better placed directly in `src/Tools/LibRASR/LabelScorer.cc`. And I would change the function signature to:
`Nn::ScaledLabelScorer* getSubScorer(Nn::ScaledLabelScorer& labelScorer, size_t index = 0)`
i.e.
- Get rid of the static
- Make the index non-optional (currently you anyway never check if it really contains a value before you access it which is unsafe)
- Change the pointer type to a ref for `labelScorer`

src/Python/Search.hh
```diff
@@ -38,10 +39,23 @@ struct TracebackItem {
 
 typedef std::vector<TracebackItem> Traceback;
 
+// Return the lanuage model
+static Lm::ScaledLanguageModel* getLanguageModel(Speech::ModelCombination& modelCombination) {
+    return modelCombination.languageModel().get();
+}
+
+// Return the label scorer
+static Nn::ScaledLabelScorer* getLabelScorer(Speech::ModelCombination& modelCombination, std::optional<size_t> index = 0) {
```
SimBe195: Here you can also get rid of `std::optional`.

src/Nn/LabelScorer/CombineLabelScorer.cc
```diff
@@ -69,6 +69,10 @@ CombineLabelScorer::CombineLabelScorer(Core::Configuration const& config, ModelC
     }
 }
 
+Core::Ref<ScaledLabelScorer> CombineLabelScorer::getSubScorer(size_t index) const {
+    return scorers_[index];
```
SimBe195, line 112: Make sure that the index is not out-of-bounds.

src/Python/LabelScorer.hh
```diff
@@ -19,10 +19,37 @@
...
+    Core::Ref<Nn::LabelScorer> wrappedLabelScorer = labelScorer->labelScorer();
+
+    auto* combineScorer = dynamic_cast<Nn::CombineLabelScorer*>(wrappedLabelScorer.get());
+    if (combineScorer) {
+        return combineScorer->getSubScorer(*index).get();
+    }
+
+    auto* encoderDecoderScorer = dynamic_cast<Nn::EncoderDecoderLabelScorer*>(wrappedLabelScorer.get());
+    if (encoderDecoderScorer) {
```
SimBe195: Maybe require that `index == 0` for variants that only have one sub-scorer.

src/Python/LabelScorer.hh
```diff
@@ -19,10 +19,37 @@
...
+#include <Nn/LabelScorer/EncoderDecoderLabelScorer.hh>
 #include <Nn/LabelScorer/LabelScorer.hh>
+#include <Nn/LabelScorer/ScaledLabelScorer.hh>
 
 namespace py = pybind11;
 
+// Return the sub-scorer that is wrapped by the given label scorer
+// If the label scorer is a CombineLabelScorer, return the sub-scorer at the specified index
+static Nn::ScaledLabelScorer* getSubScorer(Nn::ScaledLabelScorer* labelScorer, std::optional<size_t> index = 0) {
+    Core::Ref<Nn::LabelScorer> wrappedLabelScorer = labelScorer->labelScorer();
```
SimBe195: You can use `auto` here.

src/Python/Search.hh
```diff
@@ -38,10 +39,23 @@ struct TracebackItem {
 
 typedef std::vector<TracebackItem> Traceback;
 
+// Return the lanuage model
```
SimBe195: Again, I would move these functions to `src/Tools/LibRASR/Search.cc` because these are just pybind helpers and unused otherwise.

SimBe195 (changes requested):

SimBe195 (approved):

src/Python/LabelScorer.hh
```diff
@@ -19,7 +19,11 @@
 #include <pybind11/pybind11.h>
 #include "Nn/LabelScorer/Types.hh"
 
+#include <Nn/LabelScorer/CombineLabelScorer.hh>
+#include <Nn/LabelScorer/CtcPrefixLabelScorer.hh>
+#include <Nn/LabelScorer/EncoderDecoderLabelScorer.hh>
 #include <Nn/LabelScorer/LabelScorer.hh>
+#include <Nn/LabelScorer/ScaledLabelScorer.hh>
```
SimBe195: All these includes can be cleaned up again.

src/Python/Search.hh
```diff
@@ -18,6 +18,7 @@
 
 #include <Flf/LatticeHandler.hh>
 #include <Flf/Lexicon.hh>
+#include <Nn/LabelScorer/ScaledLabelScorer.hh>
```
SimBe195: This include too.

src/Python/LabelScorer.hh
larissakl: Yes, I'm already on it!

larissakl:
@SimBe195 Can we merge now or do you have anything else?

PR #238, opened 2026-06-11
Add <optional> tag to Bliss::CorpusParser

curufinwe:
Next step is to add a shortcut for optional words:
```
<orth>Prefix <optional>middle </optional>suffix </orth>
```

is equivalent to:
```
<orth>Prefix <alternatives><orth>middle </orth><orth/></alternatives>suffix </orth>
```


SimBe195 (approved):

larissakl (approved):

src/Bliss/CorpusParser.cc
```diff
@@ -248,6 +252,13 @@ void OrthographyElement::appendAlternatives(std::vector<Orthography> const& alte
     product_.appendAlternative(alternatives);
 }
 
+void OrthographyElement::appendOptional(Orthography const& orth) {
+    std::vector<Orthography> alternatives;
+    alternatives.push_back(orth);
+    alternatives.push_back(Orthography());
```
hannah220: This `alternatives.push_back(Orthography());` is for empty orth?
Maybe check that the passed `orth` is not empty, so that we don't create twice.

PR #241, opened 2026-07-08
Add FullContextOnnxLabelScorer

larissakl:
Adds a `FullContextOnnxLabelScorer`, that forwards either the input feature at the current timestep or the full encoder state sequence, together with the complete label history through a single ONNX model. It's structurally analogous to `FixedContextOnnxLabelScorer`, but the history is never truncated.

I moved the `SeqStepScoringContext` to `ScoringContexts.[hh|cc]` again to reuse it for this label scorer.


src/Nn/LabelScorer/FullContextOnnxLabelScorer.hh
```diff
@@ -0,0 +1,130 @@
+/** Copyright 2025 RWTH Aachen University. All rights reserved.
```
SimBe195:
```suggestion
/** Copyright 2026 RWTH Aachen University. All rights reserved.
```

src/Nn/LabelScorer/FullContextOnnxLabelScorer.cc
```diff
@@ -0,0 +1,421 @@
+/** Copyright 2025 RWTH Aachen University. All rights reserved.
```
SimBe195:
```suggestion
/** Copyright 2026 RWTH Aachen University. All rights reserved.
```

src/Nn/LabelScorer/CMakeLists.txt
```diff
@@ -21,8 +21,8 @@ target_sources(
 if(${MODULE_ONNX})
     target_sources(
         RasrNn
-        PRIVATE FixedContextOnnxLabelScorer.cc NoContextOnnxLabelScorer.cc
         PRIVATE FixedContextOnnxLabelScorer.cc
+        FullContextOnnxLabelScorer.cc
```
SimBe195: Formatting is off here.

src/Nn/LabelScorer/FullContextOnnxLabelScorer.cc
```diff
@@ -0,0 +1,421 @@
...
+
+namespace Nn {
+
+const Core::ParameterInt FullContextOnnxLabelScorer::paramStartLabelIndex(
+        "start-label-index",
+        "Initial history in the first step is filled with this label index.",
+        0);
+
+const Core::ParameterInt FullContextOnnxLabelScorer::paramInitialHistoryLength(
+        "initial-history-length",
```
SimBe195: Do we need this? Is there ever a case where we repeatedly feed the same start label? I would either remove this or convert the `ParameterInt` to `ParameterIntVector` so that users can pass a proper list of start labels like in the StateManaged LabelScorer if we want the extra flexibility.

src/Nn/LabelScorer/FullContextOnnxLabelScorer.cc
```diff
@@ -0,0 +1,421 @@
...
+                "history",
+                Onnx::IODirection::INPUT,
+                false,
+                {Onnx::ValueType::TENSOR},
+                {Onnx::ValueDataType::INT32},
+                {{-1, -2}}},  // [B, historyLength]
+        Onnx::IOSpecification{
+                "history-size",
+                Onnx::IODirection::INPUT,
+                true,
```
SimBe195: Does it make sense to have this as optional? I would think that this is always required in practice.

src/Nn/LabelScorer/FullContextOnnxLabelScorer.cc
```diff
@@ -0,0 +1,421 @@
...
+
+    // Cast scoring contexts to concrete types
+    std::vector<SeqStepScoringContextRef> seqStepScoringContexts;
+    seqStepScoringContexts.reserve(scoringContexts.size());
+    for (auto const& scoringContext : scoringContexts) {
+        seqStepScoringContexts.push_back(Core::ref(dynamic_cast<SeqStepScoringContext const*>(scoringContext.get())));
+    }
+
+    /*
+     * Collect all requests that have the same history length. If a per-frame input-feature is
```
SimBe195: Why do you group by history length?

SimBe195 (changes requested):

src/Nn/LabelScorer/FullContextOnnxLabelScorer.cc
larissakl: You're right, I just removed it

src/Nn/LabelScorer/FullContextOnnxLabelScorer.cc
larissakl: Ok, I thought about this again and had a short discussion with Robin. And I agree, making the history-length tensor non-optional and then rely on the model to handle it correctly should be a valid assumption.

src/Nn/LabelScorer/CMakeLists.txt
larissakl: Yeah, I saw that already, the problem is that for some reason CLion formats this again like this as soon as I try to fix it...

larissakl: I fixed directly here in Github now. However I'm still wondering why this happens...

SimBe195 (approved):

PR #242, opened 2026-07-08
Add missing libboost dependency to 2026-05-19_pytorch-2.12_onnx-1.24_v1 apptainer

curufinwe:
I tried compiling RASR with the latest apptainer image and ran into two issues. One was that cmake tried to check if nvcc works by compiling a file with compute_52 (Maxwell). Cuda 13.2 does not support that anymore and Claude tracked this to the order in which we set the architectures. I.e. we need to first set the architectures variable and only then enable Cuda. The other thing that was broken is the missing boost library needed for boost::synchronized_value.


larissakl (approved):

SimBe195 (approved):

curufinwe:
FYI: I noticed in a separate branch that cppunit was also missing. I added it to this PR as well.

PR #243, opened 2026-07-08
Refactor Flow: replace usage of Core::ReferenceCounted with std::shared_ptr

curufinwe:
I wanted to test Claude Code and thought it would be a good opportunity to start some RASR refactoring work. The overall plan is to get rid of Core::ReferenceCounted as much as possible (to reduce the number of warnings we get when building RASR). To keep things reviewable I only started with usage in the Flow module (plus of course changes in other modules resulting from changes in Flow). This is a module we/I use a lot still so it's easier to test.

What do you think?


larissakl (approved):
I skimmed through it. Let's hope for the best...

src/Flow/Aggregate.cc
```diff
@@ -117,9 +117,9 @@ Core::XmlWriter& Aggregate::dump(Core::XmlWriter& o) const {
 //==========================================================================================
 
 bool AggregationNode::configure() {
-    Core::Ref<Attributes> a(new Flow::Attributes);
+    std::shared_ptr<Attributes> a(new Flow::Attributes);
```
SimBe195: In cases like these we could use `auto a = std::make_shared<Flow::Attributes>()` instead. In some places this is done already but there are still a lot more left. We can probably get rid of most if not all `new` declarations.

SimBe195 (changes requested):

SimBe195 (approved):

PR #247, opened 2026-07-16
Update SearchV2 algorithms: Remove hypotheses with non-finite scores

SimBe195:
LabelScorers like the `CtcPrefixLabelScorer` can return infinite score for impossible extensions (i.e. label-sequence requires alignments of length > T). Whenever this happens, it can crash the search during histogram pruning because the condition
```
void setLimits(Score lower, Score upper) {
    require_lt(lower, upper);
    ...
}
```
gets violated. This PR erases all hypotheses with infinite score before histogram pruning is applied.


larissakl (approved):

curufinwe:
From our meeting: change the erase(remove_if()) call to excluding infinite score hyps from min/max score computation and check if min-score > max-score to prune all hyps if all hyps are infinite.

curufinwe (approved):

PR #249, opened 2026-07-23
Update LabelSyncBeamSearch: Prune active and terminated hypotheses separately

SimBe195:
This PR changes the behavior of the label-synchronous search algorithm. Instead of jointly pruning all hypotheses, they are first separated into active and terminated ones. Terminated hypotheses are only pruned relative to other terminated hypotheses while active hypotheses are pruned relative to the overall beam. This ensures that early terminated hypotheses (e.g. for segments containing only silence or noise) don't get pruned away prematurely.


src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.hh
```diff
@@ -181,10 +197,20 @@ private:
 
     /*
      * Helper function for acoustic pruning of hypotheses. Calculates an absolute threshold based on best score + relative threshold and
-     * score histogram. Removes all extensions worse than the absolute threshold.
+     * score histogram. If given, referenceScore is used instead of the best score in hypotheses for relative pruning.
+     * Removes all elements worse than the absolute threshold.
+     */
+    template<typename Element>
+    void scorePruning(std::vector<Element>& hypotheses, Score relativeThreshold, size_t maxBeamSize, std::optional<Score> referenceScore = std::nullopt);
```
larissakl: I think the `referenceScore` parameter is never passed in any call.
And maybe it's nicer to also use `PruningParams` here instead of having `relativeThreshold` and `maxBeamSize` separately.

larissakl (approved):

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.hh
SimBe195: Yes, I use PruningParams parameters for all score-pruning functions now.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -781,7 +828,10 @@ bool TreeLabelsyncBeamSearch::decodeStep() {
     /*
      * Prune set of word-end hypotheses by max beam size and possibly also by score.
      */
-    scorePruning(wordEndExtensions_, wordEndScoreThreshold_, maxWordEndBeamSize_);
+    scorePruning(wordEndExtensions_,
```
curufinwe: **F1 — `separate` is wired into only the last two of the five gates in this algorithm, and this one is the strongest of the three it misses.**

`LexiconfreeLabelsyncBeamSearch` has four points where a candidate can die, and this PR made all four pool-aware. `TreeLabelsyncBeamSearch` has five, and only the final beam score pruning (`:871-912`) and the final max-beam pruning (`:946-965`) were touched. These three stay joint in *both* modes:

- `:651` — the pre-prune during within-word extension creation, which has no terminated reference score. `LexiconfreeLabelsyncBeamSearch.cc:516-534` got exactly that and this did not.
- `:702` — `scorePruning(withinWordExtensions_, …)`. That list genuinely mixes terminating and non-terminating candidates: `:637` sets `transitionType = SENTENCE_END` when the successor token is the sentence-end label, and `:89` derives `isActive` from precisely that.
- **this line** — `scorePruning(wordEndExtensions_, …)`.

With a single label scorer, the first two both reduce to "apply `score-threshold` jointly, on raw score, against the overall best extension" — which is the comparison the PR sets out to eliminate. It is eliminated in the lexiconfree search and not here.

This gate is worse than a threshold filter, though, because it is a **hard gate**: failing it destroys the hypothesis rather than merely deprioritising it.

1. `:733-756` builds `newBeam_`, with a freshly terminated hypothesis in it at a non-root state.
2. `:777-787` gives it a word-end extension carrying the sentence-end LM cost, `score = hyp.score + languageModel_->sentenceEndScore(hyp.lmHistory)`.
3. This line prunes `wordEndExtensions_` — jointly, in both modes.
4. `:857-864` then deletes every freshly terminated hypothesis from `newBeam_` outright, on the stated grounds that it has already been expanded to a word-end hypothesis:

```cpp
[this](LabelHypothesis const& hyp) { return not hyp.isActive and not network_->isRoot(hyp.currentState); }
```

So the word-end extension is the only route by which a newly terminated hypothesis reaches the beam. If step 3 dropped it, step 4 removes the base hypothesis too and it is gone before `separateScorePruning` at `:871` ever sees it.

And this is the tightest gate in the algorithm. `wordEndScoreThreshold_` is a *factor* on `score-threshold` (`:325`), so the example config in `doc/search_v2.rst` — `score-threshold = 5.0`, `word-end-score-threshold = 0.5` — makes it 2.5, half the threshold `separate` later applies; `max-word-end-beam-size = 20` there against `max-beam-size = 50`. The comparison is on raw score, since `WordEndExtensionCandidate::pruningScore()` returns `score` with no length normalisation, at the exact moment the terminated candidate has just paid its sentence-end LM cost while its competitors paid an ordinary word-transition score.

Two caveats, so this is not overstated: both word-end parameters default to off (`Core::Type<Score>::max` and `INT_MAX`), so this only bites when they are configured — it just happens that the documented example configures both. And I have not measured how often a hypothesis is actually lost here; this is a code-path argument, not an observed regression.

The root cause looks structural rather than accidental: `WithinWordExtensionCandidate` and `WordEndExtensionCandidate` (`TreeLabelsyncBeamSearch.hh:84-116`) never got an `isActive` field, whereas the lexiconfree `ExtensionCandidate` did — which is why that side could be finished. Same fingerprint in the declaration: `separateScorePruning` is a template over the element type in the lexiconfree header and a plain function over `std::vector<LabelHypothesis>` here, because `LabelHypothesis` is the only tree type that can answer "are you active?". Deriving `isActive` from `transitionType == SENTENCE_END` on both structs, the way `:89` already does, would let it become a template here too and cover all five gates.

**The question:** is the narrower scope deliberate? There is a reasonable case that it is — the terminated pool at any single word-end step is small, and one could read `separate` as protecting the beam rather than the candidate lists. If that is the intent, `doc/search_v2.rst:379-380` should say so, because `pruning-strategy-type` currently reads as one parameter with one meaning and it means something materially weaker here, with the strongest gate left joint.

— Claude Code

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -569,16 +656,50 @@ bool LexiconfreeLabelsyncBeamSearch::decodeStep() {
...
+     * Prune hypotheses by score according to the selected pruning strategy.
      */
-    if (not useScorePruning_.empty() and useScorePruning_.back()) {
-        auto relativeThreshold = scoreThresholds_.back();
-        if (lengthNormScale_ != 0) {
-            auto const* bestHypothesis = getBestHypothesis(newBeam_, HypothesisFilter::Any);
-            verify(bestHypothesis != nullptr);
-            relativeThreshold /= std::pow(bestHypothesis->length, lengthNormScale_);
+    size_t const finalScorerIdx = labelScorers_.size() - 1;
+    if (useScorePruning_[finalScorerIdx]) {
```
curufinwe, line 676: **F2 — this silently changes `joint`-mode results, i.e. every existing config, and is not mentioned in the PR description.**

Before this PR the final stage read:

```cpp
if (not useScorePruning_.empty() and useScorePruning_.back()) {
    auto relativeThreshold = scoreThresholds_.back();
```

It now indexes by `finalScorerIdx = labelScorers_.size() - 1` instead of taking the back element. `scoreThresholds_` is the configured `score-threshold` vector padded up to `maxBeamSizes_.size()` (`:233-241`), so `useScorePruning_.size() == scoreThresholds_.size() >= maxBeamSizes_.size()`, and it can be **longer** than `labelScorers_.size()`. Whenever it is, the two forms disagree:

- `max-beam-size = 50`, `score-threshold = 5.0 8.0`, one label scorer → the final beam was gated by `8.0` and is now gated by `5.0`. **No diagnostic at all**: the size warning at `:274-276` compares `labelScorers_.size()` against `maxBeamSizes_`, not against `scoreThresholds_`.
- `max-beam-size = 100 50`, one label scorer → the warning does fire, and the final threshold moves from `scoreThresholds_[1]` to `scoreThresholds_[0]`.

Because `useScorePruning_[i]` is per-entry, it can also flip final score pruning on or off outright rather than merely retuning it.

To be clear about what this is not: it is not a crash. `labelScorers_.size() > maxBeamSizes_.size()` hits `error()` at `:271-273` and `on-error` defaults to `immediate-exit`, so `finalScorerIdx` is always in bounds by the time `decodeStep` runs — which is also why dropping the `not useScorePruning_.empty()` guard is safe, if only by that coupling. Worth knowing that the safety rests on it.

I think the new indexing is the *correct* one — it matches `scoreThresholds_[scorerIdx]` in the intermediate loop at `:588`, and the last iteration of that loop already applies the same entry to the extensions. But then it is a bug fix worth stating, and `TreeLabelsyncBeamSearch` was left on the old form (`:872` `scoreThresholds_.back()`, `:951` `maxBeamSizes_.back()`), so the two algorithms now index differently for the same conceptual quantity. Whichever is intended, it would be good to have both on it.

— Claude Code

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -1048,36 +1151,36 @@ void TreeLabelsyncBeamSearch::scorePruning(std::vector<Element>& hypotheses, Sco
...
         return;
     }
 
     Score absoluteThreshold = upperScore;
 
     // Pruning by relative score threshold
-    if (relativeThreshold != Core::Type<Score>::max) {
-        absoluteThreshold = lowerScore + relativeThreshold;
+    if (pruningParams.relativeThreshold != Core::Type<Score>::max) {
+        absoluteThreshold = lowerScore + pruningParams.relativeThreshold;
```
curufinwe: **F3 — `referenceScore` never reaches this function, and unlike the lexiconfree copy this one would ignore it if it did.**

This is the half of @larissakl's earlier comment that is still open. The `PruningParams` half is genuinely done in `387c492a`; the never-passed-`referenceScore` half is not:

- `LexiconfreeLabelsyncBeamSearch.cc`'s `scorePruning` *does* honour it (`:936` for the everything-is-pruned early-out, `:954` for the threshold), but all three of its call sites leave it defaulted. Dead in practice.
- This function never mentions `referenceScore` at all — the line reads `lowerScore + pruningParams.relativeThreshold` where the lexiconfree version reads `referenceScore.value_or(lowerScore) + relativeThreshold`. Its three call sites (`:702`, `:831`, `:880`) also leave it defaulted, so there is no live bug — but the field sits in the struct this function takes, so a future caller that sets it gets it **silently dropped**. That is the kind of trap worth closing while the code is fresh rather than discovering later from a pruning result that makes no sense.

Only `separateScorePruning` actually reads it, in both files. Two ways out: move `referenceScore` out of the shared `PruningParams` into a parameter only `separateScorePruning` takes, so the type system stops advertising something `scorePruning` does not support; or make this function honour it the way the lexiconfree one does. The first is probably tidier given no `scorePruning` caller wants it.

— Claude Code

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.hh
```diff
@@ -130,13 +133,22 @@ private:
         Any,
     };
 
+    struct PruningParams {
+        Score                relativeThreshold;
+        size_t               maxBeamSize;
+        std::optional<Score> referenceScore = std::nullopt;
```
curufinwe, line 146: **F4 — `std::optional` is used here but `<optional>` is not included in this header.**

It compiles only transitively, through the `Nn/LabelScorer` headers. The same PR added `#include <optional>` to `TreeLabelsyncBeamSearch.hh:19` for the identical struct, so this looks like an oversight rather than a deliberate house convention — worth adding for symmetry, and because a change in what the `Nn` headers pull in would otherwise break this file for reasons that have nothing to do with it.

— Claude Code

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
```diff
@@ -782,36 +928,42 @@ void LexiconfreeLabelsyncBeamSearch::scorePruning(std::vector<Element>& hypothes
...
+    if (hypotheses.size() > pruningParams.maxBeamSize) {
         scoreHistogram_.clear();
         scoreHistogram_.setLimits(lowerScore, upperScore);
 
         for (auto const& hyp : hypotheses) {
             scoreHistogram_ += hyp.pruningScore();
         }
 
-        absoluteThreshold = std::min(absoluteThreshold, scoreHistogram_.quantile(maxBeamSize));
+        absoluteThreshold = std::min(absoluteThreshold, scoreHistogram_.quantile(pruningParams.maxBeamSize));
```
curufinwe, line 983: **F5 — `joint` and `separate` differ inside the histogram path in a way that is not documented and probably not intended.**

This histogram is filled by the loop just above, which has no finiteness filter — unlike the range-finding loop at the top of the function, which skips hypotheses with `Math::isinf(hyp.score) or hyp.score >= Core::Type<Score>::max`. So unscorable hypotheses land in the top bin and count toward `quantile(maxBeamSize)`, pulling the threshold down and over-pruning. The gate is on total `hypotheses.size() > maxBeamSize` as well.

`separateScorePruning` does neither: it filters to finite scores when filling its histograms (`:1063-1071`) and gates on `numFiniteScoreElements[poolIdx] > params->maxBeamSize` (`:1060`).

`separate` is the more correct of the two here — the older behaviour is the odd one. But the consequence is that at identical `max-beam-size` the two modes prune differently for a reason unrelated to pooling, which undercuts using `joint` as the baseline when measuring what `separate` buys. Same category, smaller: `separateScorePruning` drops unscorable elements in cases where `scorePruning` keeps them, because `scorePruning` returns early at `absoluteThreshold >= upperScore` (`:969-972`) before reaching the erase.

Cheapest fix is to add the same finiteness filter to this loop, so the two modes differ only where they are meant to.

— Claude Code

doc/search_v2.rst
```diff
@@ -371,6 +376,8 @@ index used for the search tree itself.
   un-normalized score units.
 * ``length-norm-scale``, ``max-labels-per-timestep``: same meaning and defaults as for
   ``lexiconfree-labelsync-beam-search`` above.
+* ``pruning-strategy-type``: same choices and default as for ``lexiconfree-labelsync-beam-search`` above, applied
```
curufinwe: **F6 — three gaps here, one of which can cost someone a confusing abort.**

- This entry does not repeat that `separate` requires a finite final `score-threshold`, although the tree code errors on it too (`TreeLabelsyncBeamSearch.cc:363-370`). Someone configuring only `tree-labelsync-beam-search` reads no warning and gets an immediate exit, since `on-error` defaults to `immediate-exit`.
- Neither entry states the practical cost of the feature: with per-pool limits the beam can hold up to **2×** `max-beam-size`, and in the lexiconfree search the *intermediate* beams can too, since both pools get `maxBeamSizes_[scorerIdx]` (`LexiconfreeLabelsyncBeamSearch.cc:579-582`). The lexiconfree entry implies it with "max-beam-size limits are applied per pool"; this entry says nothing about per-pool max-beam at all, even though `:951` applies `maxBeamSizes_.back()` to both pools.
- Neither mentions what the constructors warn about — `separate` with `length-norm-scale = 0` compares the pools by raw accumulated score (`LexiconfreeLabelsyncBeamSearch.cc:251-259`, `TreeLabelsyncBeamSearch.cc:329-337`). A warning that only exists in the log is easy to miss when configuring from the docs.

Worth one more sentence while here: that in `separate` mode the final score threshold becomes the only thing retiring active hypotheses, so a loose threshold makes the search run considerably longer than in `joint`.

— Claude Code

curufinwe (commented):
Full review of the diff at `387c492a` — all five files, both algorithms, read statically. No local build: compile and clang-format CI are green at this head, and nothing below turns on a build.

Flagging up front that the approval on this PR predates the current tree — `09956b6a` and `387c492a` landed after it, which is why GitHub reports `REVIEW_REQUIRED`. So this is a first look at the head, not a re-read.

Posting as comments rather than a verdict. Six findings, in the inline comments:

| | severity | where |
|---|---|---|
| **F1** | should block | `separate` covers only the last two of five pruning gates in the tree search |
| **F2** | should block | undocumented behaviour change in `joint` — the default |
| **F3** | low-medium | `PruningParams::referenceScore` set by no `scorePruning` caller, ignored outright by the tree copy |
| **F4** | low | `LexiconfreeLabelsyncBeamSearch.hh` uses `std::optional` without including it |
| **F5** | low | `joint` and `separate` differ inside the histogram path, undocumented |
| **F6** | low | doc gaps in `search_v2.rst` |

Recorded as **checked and clean**, so nothing needs re-explaining:

- **The default really is the old behaviour.** `paramPruningStrategyType` defaults to `PruningStrategyJoint` in both algorithms, so the new parameter changes nothing for existing configs. (F2 is an independent change that happens to ride along, not a consequence of the parameter.)
- **The one-pass compaction in `separateScorePruning` is correct.** `writeIt <= readIt` throughout, no reallocation, moves only when the iterators differ, single tail erase. No iterator or reference invalidation, and no index into one vector used after the other is compacted — `baseHypIndex` points into `beam_`, which compacting `extensions_` never touches, and the scoring-context rebuild runs afterwards over a valid prefix. The two copies of the function are identical apart from the element type; I diffed them line by line.
- **Neither pool can be emptied by score pruning at the same time**, so `newBeam_` cannot go empty that way: `pruneAll` for the active pool needs the overall-best reference to beat every active score, which requires the best to be terminated — and then the terminated pool's own reference equals its minimum.
- **The unsound-configuration guard is a genuine hard stop**, not just a log line: `on-error` defaults to `ErrorActionImmediateExit` (`src/Core/Component.cc:91-95`), so `separate` without a finite final `score-threshold` aborts rather than looping. Worth saying because it makes that part of `af8a22cd` sound.
- **`ExtensionCandidate::isActive` is equivalent to the `transitionType != SENTENCE_END` it replaces** — both are set in the same `tokenIdx == sentenceEndLabelIndex_` branch. Redundant state, but deliberately so, since it lets the templated `matchesHypothesisFilter` serve both element types.
- **Omitting `isActive` from the lexiconfree recombination context is safe**, because a terminated hypothesis always has `currentToken == sentenceEndLabelIndex_` and an active one never does, so `currentToken` already separates the pools. (The tree version needs it, and has it, because a terminated hypothesis there sits at a root state an active one can also occupy.)

Two things I could not close, stated as open rather than passed:

- No decoding run, so F1 is a reading of the code path, not an observed regression. How often it bites depends on `word-end-score-threshold` and `max-word-end-beam-size`.
- In `separate` mode terminated hypotheses stay in the beam for many more steps and re-enter `recombination()` on each one, where `verify(not hyp.trace->sibling)` assumes a hypothesis is recombined at most once. I traced the ordering — carried-over terminated hypotheses are pushed into `newBeam_` before fresh extensions, so a carried-over hypothesis is always the `inserted == true` side of a collision — and I believe it is safe, but I could not rule it out by construction.

One behavioural note that is not a defect but is worth a sentence in the docs: with per-pool beam limits, terminated hypotheses can no longer displace active ones, so in `separate` mode the *only* thing that retires an active hypothesis is the final score threshold. The `error()` covers the no-threshold case; with a merely loose threshold, a search that used to end because the beam filled with terminated hypotheses will now run on until the threshold catches every active hypothesis or the `max-labels-per-timestep` cap fires.

— Claude Code

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
SimBe195: It's intentional that within-word pruning doesn't differentiate active and terminated hypotheses since within-word extensions can not be terminated yet. Only word-end extensions at the sentence-end lemma count as terminated. The word-end pruning concern was correct and is fixed now.

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
SimBe195: The new indexing is intentional. This anyway only affects configs where the number of score-thresholds is longer than the number of label scorers. For this case a warning has been added.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
SimBe195: Yes, this is valid. `referenceScore` was never used. I overlooked this when introducing the `PruningParams` struct. It's now respected in the function.

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.hh
SimBe195: There are lots of places in RASR where includes only happen transitively. But it's fine, I have added it.

src/Search/LexiconfreeLabelsyncBeamSearch/LexiconfreeLabelsyncBeamSearch.cc
SimBe195: Yes, the same finiteness filter has been added here.

doc/search_v2.rst
SimBe195: Comments for the increased `max-beam-size`, the `score-threshold` and the `length-norm-scale = 0` issue have been added to the documentation.

src/Search/TreeLabelsyncBeamSearch/TreeLabelsyncBeamSearch.cc
```diff
@@ -292,12 +317,26 @@ TreeLabelsyncBeamSearch::TreeLabelsyncBeamSearch(Core::Configuration const& conf
...
     }
 
-    if (scoreThresholds_.back() == Core::Type<Score>::max and wordEndScoreThreshold_ != Core::Type<Score>::max) {
-        error() << "Word-end score-threshold which is relative to the score-threshold is set, but score-threshold is not set";
-    }
     if (wordEndScoreThreshold_ != Core::Type<Score>::max) {
-        log() << "Use absolute word-end score-threshold of " << wordEndScoreThreshold_ * scoreThresholds_.back() << "; computed relative to within-word threshold " << scoreThresholds_.back() << " with factor " << wordEndScoreThreshold_;
-        wordEndScoreThreshold_ *= scoreThresholds_.back();
+        // The word-end threshold is configured as a factor relative to the final within-word threshold
+        Score const withinWordScoreThreshold = scoreThresholds_.empty() ? Core::Type<Score>::max : scoreThresholds_.back();
```
curufinwe: **Follow-up on the F2 fix — this `.back()` is the one place the new "surplus thresholds are ignored" warning is not true.**

Low severity, not a merge blocker; raising it because it is a one-line change if you are in the file anyway.

`setModelCombination` now warns that surplus `score-threshold` entries are ignored, and the final pruning stage genuinely does ignore them — it indexes `scoreThresholds_[finalScorerIdx]`. But `wordEndScoreThreshold_` is still derived here from `scoreThresholds_.back()`, so in exactly the configuration the warning describes, the word-end threshold is scaled off an entry nothing else uses:

`score-threshold = 5.0 8.0`, `max-beam-size = 50`, one label scorer, `word-end-score-threshold = 0.5`
→ `wordEndScoreThreshold_` becomes `0.5 * 8.0 = 4.0`, while the final beam pruning uses `5.0`. Presumably `0.5 * 5.0 = 2.5` was intended, so word-end pruning ends up about 60% looser than the config reads.

There is a second route into it that is worth knowing about separately, because `.back()` can be a *padded* entry rather than a configured one. `scoreThresholds_` is filled up to `maxBeamSizes_.size()` with `Core::Type<Score>::max` just above, so:

`max-beam-size = 50 20`, `score-threshold = 5.0`, one label scorer, `word-end-score-threshold = 0.5`
→ `scoreThresholds_` is `[5.0, max]`, `.back()` is `max`, and the `error()` below fires with *"score-threshold is set, but score-threshold is not set"* — which is not true, it was set — aborting the run by default, before the label-scorer-count warnings are printed.

To be clear, that abort is **pre-existing** — `387c492a` and earlier had the identical `.back()` test, so it is not a regression from this commit. But one thing did change in `a72314e8`, in the path where the error is configured away (`on-error = ignore` or `delayed-exit`): the multiply now sits in the `else`, so it is skipped and `wordEndScoreThreshold_` keeps its raw factor `0.5`, read from then on as an absolute threshold in score units. Previously it ran `wordEndScoreThreshold_ *= scoreThresholds_.back()` with `.back() == max`, overflowing to infinity — i.e. no word-end score pruning. So the outcome flipped from "no word-end pruning" to "word-end pruning roughly 5× tighter than intended". Only reachable with a non-default `on-error`, so this is a footnote rather than a finding, but it is a silent quality regression rather than the no-op it used to be.

All three fall out if the derivation moves to `setModelCombination` and uses `scoreThresholds_[finalScorerIdx]` — the same index the final stage uses. It cannot stay in the constructor, since `labelScorers_` is not known yet, which is presumably why it reads `.back()` in the first place. That makes the word-end threshold relative to the threshold actually in force, makes the error message truthful, and removes the ignore-path difference.

— Claude Code

curufinwe (approved):

PR #250, opened 2026-07-23
Add documentation for SearchV2

larissakl:
Me and my little helper wrote some documentation for the SearchV2. It's purely written from a user's perspective, and does not discuss implementation details unless absolutely necessary, so maybe this can be helpful for others to create a setup (or not, as they also have little helpers to do that for them...)


doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+
+SearchV2 is RASR's newer decoding framework for neural, end-to-end style recognizers (CTC, RNN-T/transducer,
+Attention encoder-decoder, ...). It replaces the acoustic feature scoring of the classic
+:ref:`Decoder` (``Search::SearchAlgorithm`` + ``Mm::FeatureScorer``) with a self-contained
+:ref:`label scorer <SearchV2 Label Scorers>` abstraction, so the search algorithm itself no longer needs to know
+from which model topology the scores come from.
+
+This page describes SearchV2 from a user's perspective: which search algorithms and label scorers are available, how to
+configure them, and how to wire them into an Flf network or a ``librasr`` Python session. It does not require reading
+any C++ code.
```
curufinwe: 😆 . You do mention the C++ class names a lot here though.

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+
+Overview
+--------
+
+``SearchV2`` and the classic ``Search`` framework can both be present in the same RASR build. They are selected
+per use case, not globally.
+
+* **Classic search** (``Search::SearchAlgorithm``, mostly ``AdvancedTreeSearch``) drives the search with
+  acoustic scores from a ``Mm::FeatureScorer`` (GMMs, hybrid NN/HMM). It is used by the :ref:`recognizer` Flf node
+  and by ``speech-recognizer``.
```
curufinwe: I guess you mean the speech-recognizer binary? If so I would clarify that here.

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+#. Determine which parts of a :ref:`Model Combination <Common component configuration>` the algorithm needs
+   (lexicon, acoustic model, language model, label scorer(s)) and construct them accordingly.
+#. Signal the start of a new segment/utterance.
+#. Feed feature vectors one at a time, or in batches, as they become available.
+#. Optionally trigger explicit decoding steps and query intermediate ("streaming") results.
+#. Signal the end of the segment, which finalizes the search for all fed features.
+#. Retrieve the final result, either as a single-best traceback or as a word lattice.
+
+Since features are pushed in and results are pulled out independently, the same algorithm implementation can
+be used both for offline decoding of a full corpus and for online/streaming decoding, without any change to
```
curufinwe:
```suggestion
be used both for offline/batch decoding and for online/streaming decoding, without any change to
```

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+
+Four search algorithms are currently implemented on top of SearchV2. Which one to pick depends on the type of model
+and vocabulary you use:
+
+* ``lexiconfree-timesync-beam-search``: time-synchronous decoding, no pronunciation lexicon required. Typical
+  use case: CTC / neural monotonic transducer models.
+* ``lexiconfree-labelsync-beam-search``: label-synchronous decoding, no pronunciation lexicon required. Typical
+  use case: attention encoder-decoder (AED) models and Speech LLMs.
+* ``tree-timesync-beam-search``: time-synchronous decoding over a pronunciation lexicon search tree, with a
+  word-level language model. The topology of the tree itself (HMM, CTC, RNA, AED, ...) is a separate choice,
```
curufinwe:
```suggestion
  word-level language model. The topology of the tree itself (HMM, CTC, RNA, ...) is a separate choice,
```

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+  see :ref:`Search tree types` below.
+* ``tree-labelsync-beam-search``: label-synchronous counterpart to ``tree-timesync-beam-search``, decoding over
+  a pronunciation lexicon search tree built with the ``aed`` tree builder, with a word-level language model.
+
+The "lexiconfree" searches treat every entry of the lexicon as a single output token and do not build a
```
curufinwe: I would mention that they technically still require a lexicon before this sentence, otherwise it's hard to understand why this explanation is necessary.

PS: see also the sentence in the next paragraph. Pick your favorite way of merging the two.

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+pronunciation search tree or apply a word-level transition model.
+The "tree" searches (also called lexicon-constrained) build a prefix tree from the pronunciation lexicon
+and additionally scores a language model at word ends.
+
+Lexicon requirements
+^^^^^^^^^^^^^^^^^^^^^
+
+Even though the "lexiconfree" searches do not build a lexical prefix tree, a lexicon is still required.
+Each lemma is treated as one output label and the lemma's index in the lexicon must match the corresponding
```
curufinwe:
```suggestion
Each lemma in the lexicon is treated as one output label and the lemma's index in the lexicon must match the corresponding
```

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+output index of the label scorer (e.g. the softmax index of a CTC/transducer/AED model). ``blank-label-index``,
+``silence-label-index`` and ``sentence-end-label-index`` (see below) are likewise resolved as the ``id()`` of the
+lemma marked ``special="blank"``/``"silence"``/``"sentence-end"``/``"sentence-boundary"`` respectively.
+
+There are two ways to provide such a lexicon:
+
+* A full Bliss XML lexicon (``file = /path/to/lexicon.xml.gz``, the default ``xml`` format). Here it is the
```
curufinwe: What do you want to say with the text inside the brackets?

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+      B
+      C
+
+  The disadvantage of this format is that it has no way to mark special lemmata. Every line just becomes a plain label.
+  So ``blank-label-index``, ``silence-label-index`` and ``sentence-end-label-index`` cannot be inferred from a
+  ``vocab-text``/``vocab-txt`` lexicon and must be passed explicitly as parameters to the search algorithm instead.
+
+The "tree" searches, in contrast, always need a full pronunciation lexicon (the ``xml`` format). They build the
+prefix search tree directly from its pronunciations, so the ``vocab-text``/``vocab-txt`` format does not apply
+there.
```
curufinwe:
```suggestion
prefix search tree directly from its pronunciations, so the ``vocab-text``/``vocab-txt`` format is not suitable
for them.
```

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+  cross-word context at all. A lighter-weight option for GMM or hybrid NN/HMM acoustic models when full triphone
+  modeling isn't needed. Sub-parameter: ``add-ci-transitions`` (bool, default ``false``) -- insert context-independent
+  acoustic transitions between words, useful for non-fluid (isolated-word-like) speech even when the acoustic
+  model was trained on fluent speech. Warning: this tree builder is currently experimental and has not been fully
+  tested; some edge cases or input configurations may not be handled correctly yet.
+
+Multiple label scorers and per-stage parameters
+------------------------------------------------
+
+All three algorithms can use more than one label scorer at once (e.g. an acoustic model scorer plus a separate
```
curufinwe:
```suggestion
All new search algorithms can use more than one label scorer at once (e.g. an acoustic model scorer plus a separate
```
Soon there will be a forth, it's better not to hardcode the exact number here.

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+
+.. code-block:: ini
+
+    [*.search-algorithm]
+    max-beam-size   = 2400 1200
+    score-threshold = 20.0 14.0
+
+Here the beam is pruned to 2400 hypotheses (with a score threshold of 20.0) after the first label scorer, and
+further pruned to 1200 hypotheses (score threshold 14.0) after the second. If only a single value is given, it is
+used after the (only) label scorer.
```
curufinwe: Why (only)? I think this sentence needs a bit of clarificaiton.

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+
+    [*.search-algorithm]
+    num-label-scorers = 1
+
+    [*.search-algorithm.label-scorer]
+    type              = state-managed-onnx
+    scale             = 1.0
+    ; further label-scorer specific parameters, e.g. ONNX model file, I/O tensor names, ...
+
+Every label scorer constructed this way (whatever its ``type``) is automatically wrapped in a
```
curufinwe:
```suggestion
Every label scorer constructed this way (whatever its ``type``) is automatically wrapped in an
```

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+
+``encoder-decoder`` is a glue label scorer combining a separate **encoder** (which pre-processes raw input
+features into encoded representations, without needing any scoring context) with a **decoder**, which is any
+other label scorer from this list (typically ``state-managed-onnx`` or ``stateful-onnx``) that consumes the
+encoder output. It automatically handles passing encoder output into the decoder as it becomes available.
+``encoder-only`` is the same idea without a "real" decoder: the encoder output *is* the score (wrapped in a
+trivial pass-through decoder), used when encoder and output/CTC layer are exported as a single ONNX graph.
+
+* ``encoder.type`` (``onnx``/``chunked-onnx``): which encoder implementation to use. Default ``onnx``.
+* ``encoder.onnx-model.*`` : the encoder's ONNX model, configured like ``onnx-model`` above.
```
curufinwe: I would put the exact encoder configuration in separate sub-sections as they are different objects and we expect more to appear (e.g. Torch).

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+``no-context-onnx`` or ``encoder-only``) and derives label-*synchronous* prefix scores from it, by summing over
+all the time-synchronous CTC alignments consistent with a given label prefix. This lets a CTC model's scores be
+used together with a label-synchronous search such as ``lexiconfree-labelsync-beam-search``, e.g. to combine an
+AED and a CTC model at the label level.
+
+* ``blank-label-index`` (int): index of the blank label in the wrapped CTC scorer's vocabulary. Default ``0``,
+  but should always be set explicitly to match your vocabulary.
+* ``vocab-size`` (int): number of labels in the wrapped CTC scorer's vocabulary. Default ``0``, but must be set
+  explicitly -- a default of ``0`` produces a degenerate, empty score matrix.
+* ``label-scorer.*`` : configuration of the wrapped time-synchronous CTC label scorer (nested selector).
```
curufinwe: Here we could reference the stateless onnx label scorer again.

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+    vocab-size                           = 5000
+    label-scorer.type                    = no-context-onnx
+    label-scorer.onnx-model.session.file = /path/to/ctc_head.onnx
+
+combine
+^^^^^^^^
+
+Log-linearly combines multiple sub-label-scorers into a single one, assuming all sub-scorers share the same
+label alphabet: ``combined_score = sum_i(score_i * scale_i)``. The combined timeframe is the maximum over the
+sub-scorers' timeframes. This is the usual way to log-linearly combine e.g. an acoustic model scorer and a
```
curufinwe: Since when is this the usual way?

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+    scorer-2.type                    = state-managed-onnx
+    scorer-2.scale                   = 0.3
+    scorer-2.onnx-model.session.file = /path/to/lm.onnx
+
+transition
+^^^^^^^^^^^
+
+Returns a fixed, configured score for each transition type, independent of any input features or label
+history. Useful to model e.g. label-loop penalties or blank/word/sentence-end exit penalties analogous to HMM
+transition penalties, without needing a model to produce them. Recognized transition types are
```
curufinwe: A list where you group by source type would be easier to read here.

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+
+Register the type *before* constructing the ``SearchAlgorithm``/``Configuration`` that references it. Python
+label scorers can be freely combined with native ones, e.g. wrapped in ``combine`` alongside an ONNX-based
+scorer, or used as the ``decoder`` of an ``encoder-decoder`` label scorer.
+
+* ``librasr.register_label_scorer_type(name, label_scorer_cls)`` : register ``label_scorer_cls`` (a subclass of
+  ``librasr.LabelScorer``) under ``name``, making it selectable via ``label-scorer.type = <name>`` in any config
+  from then on, for the lifetime of the process.
+* ``librasr.TransitionType`` : enum mirroring the C++ ``Nn::TransitionType`` values used elsewhere in this
+  document (``LABEL_TO_LABEL``, ``LABEL_LOOP``, ``LABEL_TO_BLANK``, ``BLANK_TO_LABEL``, ``BLANK_LOOP``,
```
curufinwe: I would link them here not repeat. Otherwise new types need to be added in multiple places.

doc/search_v2.rst
```diff
@@ -0,0 +1,1043 @@
...
+  without skip-transitions. Cross-word context is applied per phoneme, not tree-wide: only phonemes marked
+  context-dependent in the lexicon's phoneme inventory get a coarticulated root state and thus cross-word
+  diphone modeling; phonemes marked context-independent use the plain context-independent root and get no
+  cross-word context at all. A lighter-weight option for GMM or hybrid NN/HMM acoustic models when full triphone
+  modeling isn't needed. Sub-parameter: ``add-ci-transitions`` (bool, default ``false``) -- insert context-independent
+  acoustic transitions between words, useful for non-fluid (isolated-word-like) speech even when the acoustic
+  model was trained on fluent speech. Warning: this tree builder is currently experimental and has not been fully
+  tested; some edge cases or input configurations may not be handled correctly yet.
+
+Multiple label scorers and per-stage parameters
```
curufinwe, line 447: What is missing from the description of the label scorers is which transition type preset they have. I think that would be helpful to have here. Maybe you could also discuss how to add / remove scored transitions first and then introduce the presets.

curufinwe (changes requested):

doc/search_v2.rst
larissakl: mentioning C++ class names != explaining code != requiring the reader to read code :smile:

doc/search_v2.rst
larissakl: True, I actually already included the tree-labelsync-search here as we plan to merge it soon

doc/search_v2.rst
larissakl: I'll add some explanation about what happens if more/less values than label scorers are defined here

doc/search_v2.rst
larissakl: I guess you mean no-context-onnx? Yes, good idea

doc/search_v2.rst
larissakl: Ok, but I will add a note that they have to be in caps and in snake case

doc/search_v2.rst
larissakl: Please check if you are okay with how I changed it or if you would do it differently. I also already added a placeholder for the Torch encoder

doc/search_v2.rst
larissakl: Oh right, I completely forgot about that. That's a really important point though. I will add something, thank you for the reminder.

In general thanks a lot for your extensive feedback!

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+* ``encoder.chunk-size`` / ``step-size`` (int, in input features): size of and shift between consecutive
+  chunks. Default ``1`` each.
+* ``encoder.left-padding`` / ``right-padding`` (int): additional context padded onto each chunk on either side.
+  Default ``0`` each.
+* ``encoder.zero-padding`` (bool): whether to pad the first/last chunk with zeros to a uniform size. Default
+  ``false``.
+* ``encoder.window-type`` (enum): weighting window used to blend overlapping chunk outputs back together.
+  Default ``triangular``.
+* ``encoder.interpolation-mode`` (enum): how overlapping chunk outputs are interpolated. Default
+  ``no-interpolation``.
```
curufinwe: What are other possible values?

doc/search_v2.rst
curufinwe: I was thinking of having it as a separate section, i.e. as a sibling to label scorers, but this also works for me.

curufinwe (changes requested):
Only one minor thing remaining.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
+SearchV2 Framework
+===================
+
+``SearchV2`` is RASR's newer decoding framework for neural, end-to-end style recognizers (CTC, RNN-T/transducer,
+Attention encoder-decoder, ...). It replaces the acoustic feature scoring of the classic
```
SimBe195:
```suggestion
``SearchV2`` is RASR's newer decoding framework for neural, end-to-end style recognizers (CTC, neural transducer,
Attention encoder-decoder, Speech LLM, ...). It replaces the acoustic feature scoring of the classic
```

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+
+``SearchV2`` and the classic ``Search`` framework can both be present in the same RASR build. They are selected
+per use case, not globally.
+
+* **Classic search** (``Search::SearchAlgorithm``, mostly ``AdvancedTreeSearch``) drives the search with
+  acoustic scores from a ``Mm::FeatureScorer`` (GMMs, hybrid NN/HMM). It is used by the :ref:`recognizer` Flf node
+  and by the ``speech-recognizer`` binary.
+* **SearchV2** (``Search::SearchAlgorithmV2``) drives the search by pulling scores directly from one or more
+  :ref:`label scorers <SearchV2 Label Scorers>` (typically wrapping an ONNX model). It is used by the
+  :ref:`recognizer-v2` Flf node and by the ``librasr`` Python bindings.
```
SimBe195: Maybe explain the difference that the LabelScorer is context-aware and can be forwarded during the search process based on hypothesis context while the FeatureScorer assumes that the scores are purely based on acoustic feature input.

doc/search_v2.rst
larissakl: Ah no, I'm also fine with putting it in a separate section and add a reference pointing to it. This is actually cleaner.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+and various label scorers for different models.
+
+Core workflow
+-------------
+
+Every ``SearchV2`` algorithm implements the ``Search::SearchAlgorithmV2`` interface and is driven the same way,
+whether it is invoked by the :ref:`recognizer-v2` node or by Python:
+
+#. Determine which parts of a :ref:`Model Combination <Common component configuration>` the algorithm needs
+   (lexicon, acoustic model, language model, label scorer(s)) and construct them accordingly.
```
SimBe195: The term "acoustic model" here could be confusing to readers. This is what its called internally RASR but in the new searches we only use it for state tying and tree topology.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+#. Signal the start of a new segment/utterance.
+#. Feed feature vectors one at a time, or in batches, as they become available.
```
SimBe195: "in chunks" rather than "in batches".

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+lexiconfree-timesync-beam-search
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A simple time-synchronous beam search without a pronunciation lexicon, word-level language model or transition model.
+Intended for open-vocabulary decoding with CTC and neural transducer style or similar models. Supports
+an optional blank label (for CTC/transducer) and optional silence and sentence-end labels.
+
+* ``max-beam-size`` (int list): maximum number of hypotheses kept in the beam. Pruning is applied after every
+  label scorer in the pipeline (see :ref:`SearchV2 Label Scorers`), so one value is expected per configured
+  label scorer. Default: unset, i.e. no beam-size pruning is applied.
```
SimBe195: "Default: unset, i.e. no beam-size pruning is applied" is wrong. The user has to provide as many max-beam-size values as there are label scorers, otherwise the search will throw an error. Beam-size pruning always has to be applied as a fallback.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+  this amount. Also applied once per label scorer. Default: unset, i.e. no score-based pruning is applied.
+* ``num-histogram-bins`` (int): number of bins used for histogram pruning (minor effect on results/speed). Default ``100``.
+* ``blank-label-index`` (int): lexicon index of the blank label. Inferred automatically from a lemma with
+  ``special="blank"`` if present; if neither is set, blank handling is disabled. Default: disabled.
+* ``silence-label-index`` (int): lexicon index of the silence label, inferred from ``special="silence"`` if unset.
+  Default: disabled.
+* ``sentence-end-label-index`` (int): lexicon index of the sentence-end label, inferred from
+  ``special="sentence-end"``/``special="sentence-boundary"`` if unset. Default: disabled.
+* ``collapse-repeated-labels`` (bool): collapse repeated emission of the same label into a single output
+  (typical for CTC-style label loops). Default ``false``.
```
SimBe195: For this parameter I have seen some confusion from users before. Maybe we should expand the documentation for this one and add examples that show what effect this has.

curufinwe (approved):

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+* ``recombination-mode`` (``on``/``off``): recombine hypotheses that share the same label scorer state
+  (keeping only the best), like word-history recombination in the classic search. Default ``on``.
```
SimBe195: You could add some explanation here. If you turn this off, redundant hypotheses will stay alive but you save the recombination cost itself. So you should only turn this off if recombination will never happen anyway (e.g. full alignment-level context) and/or the recombination itself is very expensive (e.g. when you have a large vocabulary and no score-pruning enabled, resulting in a ton of extension candidates).

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A label-synchronous beam search, again without a pronunciation lexicon, word-level language model or
+transition model. Hypotheses are terminated by an explicit sentence-end symbol rather than by running out of
+input frames. Its main purpose is open-vocabulary search with attention encoder-decoder (AED) or similar
+models.
+
+* ``max-beam-size`` (int list), ``score-threshold`` (float list), ``num-histogram-bins`` (int),
+  ``recombination-mode``, ``log-stepwise-statistics``, ``cache-cleanup-interval``: same meaning as for
+  ``lexiconfree-timesync-beam-search`` above.
```
SimBe195: I would add something for the score-thresholds. It should be mentioned how they interact with the length normalization.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+    score-threshold         = 5.0
+    length-norm-scale       = 1.0
+    max-labels-per-timestep = 1.2
+
+tree-timesync-beam-search
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A time-synchronous beam search that decodes over a prefix tree built from the pronunciation lexicon. Similar
+in structure to the classic tree search, but scored via label scorer(s) instead of a feature scorer. A
+language model score is added at word ends, to disable it, don't set a file or type for the LM and set its scale
```
SimBe195, line 278: I would add explicitly that if no LM is configured, RASR will still use a zerogram as default.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+* ``max-beam-size`` (int list): maximum number of *within-word* hypotheses in the beam, one value per label scorer.
+  Default: unset, i.e. no beam-size pruning is applied.
+* ``max-word-end-beam-size`` (int): maximum number of *word-end* hypotheses kept. If unset, word-end
+  hypotheses are pruned together with within-word hypotheses using the same global beam. Default: unset.
+* ``score-threshold`` (float list): score-based pruning of *within-word* hypotheses, one value per label scorer.
+  Default: unset, i.e. no score-based pruning is applied.
+* ``word-end-score-threshold`` (float): score-based pruning of *word-end* hypotheses, relative to
+  ``score-threshold``. If unset, word-end hypotheses use global score pruning. Default: unset.
+* ``num-histogram-bins`` (int): see above. Default ``100``.
+* ``collapse-repeated-labels`` (bool): see above. Default ``false``.
```
SimBe195: Users might be confused how this parameter interacts with the search tree topology.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+    [*.acoustic-model]
+    ; allophones / state-tying configuration, see "Common component configuration"
+
+    [*.language-model]
+    type                     = ARPA
+    file                     = /path/to/lm.gz
+    scale                    = 1.5
+
+The search tree is cached on disk next to the lexicon after it is built the first time, so subsequent runs
+with the same lexicon/acoustic-model configuration start up faster.
```
SimBe195: Not by default; the cache archive file has to be configured. In the example above, the tree would not be cached.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+
+All parameters shown in the sections below are relative to that same ``search-algorithm`` selector, e.g. the
+full path of ``max-beam-size`` is ``*.search-algorithm.max-beam-size``.
+
+lexiconfree-timesync-beam-search
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A simple time-synchronous beam search without a pronunciation lexicon, word-level language model or transition model.
+Intended for open-vocabulary decoding with CTC and neural transducer style or similar models. Supports
+an optional blank label (for CTC/transducer) and optional silence and sentence-end labels.
```
SimBe195, line 188: For all searches I would add an outline of score-contribution, pruning and recombination steps. It doesn't need to include optimizations such as on-the-fly pre-pruning but users should get an idea which sets of extensions/hypotheses are pruned in which order.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+Built-in label scorer types include:
+
+* ``no-context-onnx``, ``fixed-context-onnx``, ``stateful-onnx``, ``state-managed-onnx``: forward
+  features (and, depending on type, label history/hidden state) through an ONNX model.
+* ``encoder-decoder`` / ``encoder-only``: wrap a separate encoder (see ``encoder`` sub-selector) that
+  pre-processes features, combined with a decoder label scorer (or no decoder, for encoder-only models).
+* ``ctc-prefix``: wraps a time-synchronous CTC scorer and derives label-synchronous prefix scores from it
+  (useful with ``lexiconfree-labelsync-beam-search``).
+* ``combine``: log-linearly combines multiple sub-label-scorers into one (e.g. AM + LM), configured via nested
+  selectors.
```
SimBe195: Maybe add that most of the time you should prefer to use multiple label scorers and do intermediate pruning instead of a single combine scorer.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+``encoder.type = torch``. TODO.
+
+ctc-prefix
+^^^^^^^^^^^
+
+Wraps a time-synchronous CTC scorer (any other label scorer producing per-frame CTC-style scores, e.g.
+``no-context-onnx`` or ``encoder-only``) and derives label-*synchronous* prefix scores from it, by summing over
+all the time-synchronous CTC alignments consistent with a given label prefix. This lets a CTC model's scores be
+used together with a label-synchronous search such as ``lexiconfree-labelsync-beam-search``, e.g. to combine an
+AED and a CTC model at the label level.
```
SimBe195: Add a reference to the paper describing in more detail how the algorithm works.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+    config.set("*.search-algorithm.type", "lexiconfree-timesync-beam-search")
+    config.set("*.search-algorithm.max-beam-size", "12")
+    config.set("*.lexicon.file", "/path/to/lexicon.xml.gz")
+
+* ``Configuration()`` : create an empty configuration (default selection root is ``lib-rasr``, so both
+  ``*.foo`` and ``lib-rasr.foo`` style keys apply to it).
+* ``config.set(name, value="true")`` : set a single configuration key, e.g. ``config.set("*.search-algorithm.type", "tree-timesync-beam-search")``.
+* ``config.set_from_file(path)`` : load a ``.ini``-style RASR config file, same as ``--config=path`` for the
+  command-line tools. Returns ``True`` on success.
+* ``config[name]`` : read back a resolved value (``__getitem__``), returns ``None`` if unset.
```
SimBe195: Mention that this returns the config values as strings and users might have to post-process them.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+    for feature_chunk in feature_stream:
+        search.put_feature(feature_chunk)      # single feature, shape [F] or [1, F]
+        # or: search.put_features(feature_chunk)  # multiple features, shape [T, F] or [1, T, F]
+
+        # optional: peek at an intermediate (possibly unstable) result while streaming
+        partial = search.get_current_best_traceback()
+    search.finish_segment()            # signal that the segment is complete
+
+    best = search.get_current_best_traceback()      # final single-best result
+    stable = search.get_common_prefix()              # stable prefix shared by all current hypotheses
```
SimBe195: It doesn't make sense to call this after `finish_segment`. At that point, everything is stable. This should be moved to the loop above.

doc/search_v2.rst
```diff
@@ -0,0 +1,1158 @@
...
+* ``get_current_n_best_list(n)`` : return a list of up to ``n`` ``Traceback`` objects.
+* ``recognize_segment(features)`` : convenience method equivalent to ``enter_segment`` + ``put_features`` +
+  ``finish_segment`` + ``get_current_best_traceback`` in one call. ``features`` has shape ``[T, F]`` or ``[1, T, F]``.
+* ``recognize_segment_n_best(features, n)`` : same, but returns ``get_current_n_best_list(n)``.
+* ``model_combination()`` : return the underlying ``ModelCombination`` (see below) for accessing/adjusting
+  scales at runtime.
+
+**Traceback / TracebackItem:** a ``Traceback`` is a plain Python list of ``TracebackItem``, each with
+read/write attributes ``lemma`` (str), ``am_score`` (float), ``lm_score`` (float), ``start_time`` (int) and
+``end_time`` (int, both in feature-frame units). ``str(item)`` gives just the lemma.
```
SimBe195: A few more details could be added concerning the traceback item attributes. For example that in label-sync search `start_time` and `end_time` don't refer to timeframes but the position in the output token sequence. And that `am_score` and `lm_score` are the scaled scores. And `lm_score` refers to the word-level LM. If an LM is configured as a label scorer, its scores will contribute to `am_score` instead.

SimBe195 (changes requested):

doc/search_v2.rst
larissakl: Yeah, that's true, but how would you put it then? Use remove the term "acoustic model" here?

doc/search_v2.rst
larissakl: Good catch!

doc/search_v2.rst
SimBe195: I would write down the sub-components of the AcousticModel mode, i.e., "state tying" and "transition model".

doc/search_v2.rst
larissakl: I will add something, but we will need to change a part of the explanation for the labelsync search anyway when we add the separate pruning of active and terminated hyps. I think I'll take care of this when I implement this feature for the tree-labelsync-search

doc/search_v2.rst
larissakl: Agree, but as I already commented above, we will need such an explanation for the new pruning concept of the labelsync-searches. We could then of course also explain it for the timesync-searches. If you are okay, I would however postpone this and take care of it as soon as we have the labelsync-searches updated.

SimBe195 (approved):

PR #251, opened 2026-07-24
Update LibRASR: Add lazy logging initialization and explicit init_logging function

SimBe195:
Currently, when `librasr` is imported in Python, a `DummyApplication` is created which initializes a channel manager for logging. Since no config files can be passed to librasr at that point yet, the `DummyApplication` uses an empty config for this and only sets the encoding to "utf-8" (hardcoded). This makes it impossible to configure the channels (e.g. setting `unbuffered = yes` or similar things). This PR overhauls this mechanism. A pybind function `librasr.init_logging(config)` is added and instead of creating the channel manager immediately upon import, this is done when either `init_logging` is called or lazily when logs need to be written. This makes it possible to pass user config for the channel manager like this:
```
import librasr
config = librasr.Configuration()
config.set_from_file("rasr.config")
librasr.init_logging(config)
```
if `init_logging` is never called, the behavior stays the same, i.e., as soon as components start logging things, a channel manager is created with default parameters.

The PR also removes `initRASR` and `finiRASR` from src/Tools/LibRASR/LibRASR.cc which dont't seem to have a purpose (anymore) and are never called anywhere.


src/Core/Application.cc
```diff
@@ -300,6 +300,9 @@ int Application::run(const std::vector<std::string>& arguments) {
 }
 
 void Application::openLogging() {
+    if (channelManager_)
+        return;
+
```
curufinwe:
```suggestion
    if (channelManager_) {
        return;
    }

```

src/Core/Application.cc
```diff
@@ -312,6 +315,9 @@ void Application::openLogging() {
 }
 
 void Application::closeLogging(bool configAvailable) {
+    if (!channelManager_)
+        return;
+
```
curufinwe:
```suggestion
    if (not channelManager_) {
        return;
    }

```

curufinwe (changes requested):

larissakl (approved):
@SimBe195 If you have time, could you maybe directly add a short bullet point explaining this in the new documentation (should go under section "Using SearchV2 from Python" -> "Configuration object")?

curufinwe:
LGTM, but let's merge only after #250 is merged as you included it in this PR.

curufinwe (approved):

PR #253, opened 2026-07-30
Fix code-block rendering in documentation of Flf nodes

larissakl:
The code-blocks in the documentation of the Flf nodes all used a single colon instead of `::` which broke the rendering such that only the headings were visible


michelwi (approved):
merci

SimBe195 (approved):

PR #257, opened 2026-08-20
Add BLANK_EXIT transition type for LabelScorers

SimBe195:
Currently, the new tree search algorithms don't distinguish blank lemma exits; they are just classified as regular WORD_EXIT transitions (or can be marked as NONWORD manually). This PR adds a BLANK_EXIT transition type which is assigned to the tree exit corresponding to the blank lemma.

There is one thing to discuss: the tree builder doesn't add loops for the blank lemma state which means that for blank segments between words, every frame counts as a separate blank lemma emission of length 1. The intention behind this is to avoid redundant paths which will later get recombined anyway but now this means that the BLANK_EXIT penalty will be applied for every frame which is unexpected (and undesirable) behavior. The main ways to fix this are:

- conditionally add loops to the blank state in the tree based on a config parameter and expect users to enable this when using a scorer that scores BLANK_EXIT
- always add loops to the blank state in the tree and just accept the redundant paths in standard cases (here we could also add a tiny default BLANK_EXIT penalty so that in recombination the loop transition is preferred over exit + forward).

### DIFF
--- a/src/Nn/LabelScorer/TransitionTypes.hh
+++ b/src/Nn/LabelScorer/TransitionTypes.hh
@@ -37,6 +37,7 @@ enum TransitionType {
     WORD_EXIT,
     NONWORD_EXIT,
     SILENCE_EXIT,
+    BLANK_EXIT,
     SENTENCE_END,
     numTypes,  // must remain at the end
 };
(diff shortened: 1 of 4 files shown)


larissakl:
Originally, in the first version of the CtcTreeBuilder, the blank lemma state had a self-loop. I later removed it when I implemented the TreeTimesyncBeamSearch, but not because of redundant hypotheses, but because of the pruning strategy. In the loop, the hypotheses would count as within-word hypotheses and not word-end hypotheses during pruning. I remember we discussed this and of course it's a question of definition (if the word-end is at the last token of the word or the start of the next word, and if successive blanks "count" towards the previous or the next word) and I assume that the effect wouldn't be significant, but technically, I don't think that a hypothesis which is in a blank state between two words should be seen as a within-word hypothesis.

